PR #8, opened 2018-08-07
Arxivng 1042

DavidLFielding:
This PR includes significant changes to upload API, error formatting, and exception handling.

The API supports single upload request that create a new upload workspace, accepts subsequent file/package uploads, and provides summary information about the upload.

Errors and warning were originally just a message and portions were duplicated in the log (now errors) section of the response and in the files list. There is now a single list of errors that includes type [warn,fatal,info], public file path (if error comes from file check), and an error message. This seems much cleaner and allows complete freedom in how errors are displayed in the UI (at top of page or embedded in file list).

Exception handling has been improved significantly. Controller is doing better job of identifying and raising errors expected per API specification.

I started flushing out some of the other requests (remove workspace) and have added a significant number of comments while thinking about these new API requests. The delete workspace function still needs to be implemented. Mainly need to add code to remove the workspace directory.

Individual file deletion still needs to be implemented.

I will add more details as time permits.

I expect further discussion on specific errors/warning messages. We may want to combine certain error events and revise message wording based on experimentation with UI.


filemanager/process/upload.py
```diff
@@ -438,15 +450,15 @@ def check_files(self) -> None:
                 if file_name.startswith('-'):
                     # Replace dash (-) with underscore
                     new_file_name = re.sub('^-', '_', file_name)
-                    self.add_warning('We do not accept files starting with a hyphen. '
+                    self.add_warning(obj.public_filepath, 'We do not accept files starting with a hyphen. '
                                      + 'Attempting to rename \"' + file_name + '\" to \"'
                                      + new_file_name + '\".')
```
erickpeirson: Note that you don't need the ``+`` within ``( )``. Also the double-quotes don't need to be escaped inside a single-quoted string. E.g:

```python
self.add_warning(obj.public_filepath, 
                 'We do not accept files starting with a hyphen. '
                 f'Attempting to rename "{file_name}" to "{new_file_name}".')
```

filemanager/routes/upload_api.py
```diff
@@ -22,47 +22,60 @@ def service_status() -> tuple:
...
 
 
-@blueprint.route('/create', methods=['GET'])
-@authorization.scoped('read:upload')
-def create_one() -> tuple:
-    """Create upload workspace. Create and return unique upload identifier."""
-    data, status_code, headers = upload.create_upload()
-    return jsonify(data), status_code, headers
+@blueprint.route('/', methods=['POST'])
+@authorization.scoped('write:upload')
```
erickpeirson: Looks like in [``arxiv.users.auth.scopes``](https://github.com/cul-it/arxiv-auth/blob/develop/users/arxiv/users/auth/scopes.py) we have ``[zone]:[action]``. May want to invert these (e.g. ``upload:write``). When you're ready, we can introduce them in the ``arxiv.users`` package.

filemanager/process/upload.py
```diff
@@ -419,16 +431,16 @@ def check_files(self) -> None:
                 if re.search(r'[^\w\+\-\.\=\,]', file_name):
                     # Translate bad characters
                     new_file_name = re.sub(r'[^\w\+\-\.\=\,]', '_', file_name)
-                    self.add_warning("We only accept file names containing the characters: "
+                    self.add_warning(obj.public_filepath, "We only accept file names containing the characters: "
                                      + "a-z A-Z 0-9 _ + - . , =")
-                    self.add_warning('Attempting to rename ' + file_name
+                    self.add_warning(obj.public_filepath, 'Attempting to rename ' + file_name
```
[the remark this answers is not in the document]
DavidLFielding: I've started changing these old string cats over to f-string as I'm working in the area. I know I did a few during this sprint. I'll try to make a dedicated pass to purge these gremlins from the file management service.

filemanager/process/upload.py
DavidLFielding: Thanks. Old habits die hard. I didn't realize the '+' was optional inside (). I do recall reading the double quote " doesn't need to be escape inside single quote. I will add these to my cleanup list.

DavidLFielding: I'm surprised pylint doesn't complain at me for these...

filemanager/routes/upload_api.py
DavidLFielding: OK

DavidLFielding:
It doesn't look like anyone else will have time to look at this so I will go ahead and merge.

PR #9, opened 2018-08-09
ARXIVNG-1076 implemented authorization checks with arxiv.users

erickpeirson:
This brings the API spec up to date with [``arxiv.users==0.1.0rc1``](https://github.com/cul-it/arxiv-auth/releases/tag/0.1.0rc1). Specifically:

- Using ``arxiv.users.auth.decorators.scoped`` to enforce authorization scope.
- Using constants from ``arxiv.users.auth.scopes``
- Introduces ``is_owner()`` authorization callback in ``routes.upload_api``
- Adds simple caching using the ``flask.g`` global object to ``services.uploads.retrieve``
- Minor change to the signature of ``controllers.upload.upload`` to get the correct user ID from the authorized request.
- Updated existing tests to be consistent with the above changes.

@DavidLFielding In one of the tests you were checking the response status code when an auth token was not present, or was blank. Currently ``arxiv.users`` treats that situation as 401 Unauthorized. I updated the test accordingly. If it seems like that is the wrong status code, we can look at whether a change should be made to the ``arxiv.users`` package.

Oh, and I also added a helper script for generating auth tokens, ``generate_token.py`` (see README.md for usage instructions). This should probably be added to the ``arxiv.users`` package at some point.


erickpeirson:
@DavidLFielding I updated the PR based on #10 . One test ``TestUploadAPIRoutes.test_delete_upload_workspace`` is failing, but it looks like that's pending some additional work on workspace deletion.

erickpeirson:
@DavidLFielding Thanks! 

Depends on where in the app you want it. In the routes, you'll have access to the user at ``request.session.user`` (ID at ``request.session.user.user_id``, but be sure to check that ``request.session.user is not None``). The simplest thing might be to grab it like that in the route, and then pass the user or user ID as a parameter to the controller.

PR #12, opened 2018-08-12
Arxivng 1075

DavidLFielding:
This PR is isolated to Delete Upload Workspace request.

This current implementation stashes a copy of source.log into deleted_workspace_logs. Then is removes the entire upload workspace directory. No backups are made.

DB entry for upload workspace is retained and set to 'DELETED'.

If something else comes to mind I will try to update this description before Monday.

### DIFF
--- a/filemanager/controllers/upload.py
+++ b/filemanager/controllers/upload.py
@@ -55,6 +55,7 @@ UPLOAD_DELETED_FILE = {'deleted file'}
 UPLOAD_DELETED_WORKSPACE = {'deleted workspace'}
 UPLOAD_FILE_NOT_FOUND = {'file not found'}
 UPLOAD_DELETED_ALL_FILES = {'deleted all files'}
+UPLOAD_WORKSPACE_NOT_FOUND = {'workspcae not found'}
 
 # upload status codes
 INVALID_UPLOAD_ID = {'reason': 'invalid upload identifier'}
@@ -143,6 +144,10 @@ def delete_workspace(upload_id: int) -> Response:
             # Update database (but keep around) for historical reference. Does not
             # consume very much space. What about source log?
             # Create Upload object
+            if upload_obj.state == 'DELETED':
+                logger.info(f"{upload_id}: Workspace has already been deleted:"
+                            f"current state is '{upload_obj.state}'")
+                raise NotFound(UPLOAD_WORKSPACE_NOT_FOUND)
 
             uploadObj = filemanager.process.upload.Upload(upload_id)
 
@@ -162,7 +167,7 @@ def delete_workspace(upload_id: int) -> Response:
         logger.error(f"{upload_obj.upload_id}: Delete workspace request failed ")
         raise InternalServerError(CANT_DELETE_FILE)
     except NotFound as nf:
-        logger.info(f"{upload_id}: DeleteWorkspace: '{nf}'")
+        logger.info(f"{upload_id}: Delete Workspace: '{nf}'")
         raise
     except Exception as ue:
         logger.info("Unknown error in delete workspace. "
--- a/filemanager/controllers/upload.py
+++ b/filemanager/controllers/upload.py
@@ -52,10 +52,9 @@ UPLOAD_DB_ERROR = {'unable to create/insert new upload workspace into database'}
 UPLOAD_IO_ERROR = {'encountered an IOError'}
 UPLOAD_UNKNOWN_ERROR = {'unknown error'}
 UPLOAD_DELETED_FILE = {'deleted file'}
+UPLOAD_DELETED_WORKSPACE = {'deleted workspace'}
 UPLOAD_FILE_NOT_FOUND = {'file not found'}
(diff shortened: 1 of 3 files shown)


filemanager/controllers/upload.py
```diff
@@ -52,10 +52,10 @@
 UPLOAD_IO_ERROR = {'encountered an IOError'}
 UPLOAD_UNKNOWN_ERROR = {'unknown error'}
 UPLOAD_DELETED_FILE = {'deleted file'}
+UPLOAD_DELETED_WORKSPACE = {'deleted workspace'}
 UPLOAD_FILE_NOT_FOUND = {'file not found'}
 UPLOAD_DELETED_ALL_FILES = {'deleted all files'}
-
-UPLOAD_DELETE_WORKSPACE = {'reason': 'workspace scheduled for deletion.'}
+UPLOAD_WORKSPACE_NOT_FOUND = {'workspcae not found'}
```
erickpeirson: Typo

filemanager/controllers/upload.py
```diff
@@ -93,49 +93,81 @@
...
 
             # Update database (but keep around) for historical reference. Does not
             # consume very much space. What about source log?
+            # Create Upload object
+            if upload_obj.state == 'DELETED':
+                logger.info(f"{upload_id}: Workspace has already been deleted:"
+                            f"current state is '{upload_obj.state}'")
+                raise NotFound(UPLOAD_WORKSPACE_NOT_FOUND)
+
+            uploadObj = filemanager.process.upload.Upload(upload_id)
```
erickpeirson: Having both names ``upload_obj`` and ``uploadObj`` is a bit confusing; took a minute to see that ``upload_obj`` represents the database entry, and ``uploadObj`` represents an instance of the upload workspace manager. Maybe ``upload_data`` and ``upload_manager``?

filemanager/controllers/upload.py
```diff
@@ -93,49 +93,81 @@
...
             # Actually remove entire workspace directory structure. Log
-            # something to global log since source log is being removed!
+            # everything to global log since source log is being removed!
 
             # Initiate workspace deletion
 
             # Update database (but keep around) for historical reference. Does not
             # consume very much space. What about source log?
+            # Create Upload object
+            if upload_obj.state == 'DELETED':
```
erickpeirson: Might be something for another PR, but consider defining state flags as class attributes. E.g. 

```python
class Upload(Data):
    DELETED = 'DELETED'
```

The advantages of this are that it improves self-documentation of the ``Upload`` class, and it also ensure that you will get feedback if you spell it wrong. E.g.  if ``upload_obj.state == 'DEELTED':`` (note typo) will fail silently, whereas ``upload_obj.state == Upload.DEELTED`` will cause an ``AttributeError``.

filemanager/controllers/upload.py
```diff
@@ -145,11 +177,10 @@ def delete_workspace(upload_id: int) -> Response:
     # API doesn't provide for returning errors resulting from delete.
     # 401-unautorized and 403-forbidden are handled at routes level.
     # Add 400 response to openapi.yaml
-    if upload_id:
-        raise NotImplemented(REQUEST_NOT_IMPLEMENTED)
 
-    #status_code = status.HTTP_204_NO_CONTENT
-    #return UPLOAD_DELETE_WORKSPACE, status_code, {}
+    response_data = UPLOAD_DELETED_WORKSPACE  # Get rid of pylint error
```
erickpeirson: Looks like ``UPLOAD_DELETED_WORKSPACE`` is a ``set`` (line 55). Do you mean for these to be ``dict``? E.g. 

```python
UPLOAD_DELETED_WORKSPACE = {'details': 'deleted workspace'}
```

filemanager/controllers/upload.py
DavidLFielding: Yes, I've been thinking this is a little confusing. I will try to clear this up. I wonder if 'upload_workspace' is closer to what the Upload object represents.

filemanager/controllers/upload.py
DavidLFielding: I've cleaned this up. This seemed to go through some evolution when exception handlers were added and I believe I've made these messages more consistent in the file management code.

filemanager/controllers/upload.py
erickpeirson: That makes sense to me. Might consider even renaming the class (``UploadWorkspace`` eg) at some later point

erickpeirson:
This is looking great!

filemanager/controllers/upload.py
DavidLFielding: I've replaced status strings with Upload class attributes. Much cleaner now. Pushed to github.

I will rename other Upload class to UploadWorkspace in a future PR. This will help eliminate any remaining confusion.

DavidLFielding:
I'm thinking to merge this PR and start to setup next branch for work on release/lock requests. I'm happy with some of the cleanup which has been on my to do list after getting confused myself at times.

PR #14, opened 2018-08-16
Arxivng 1077 1078 1079

DavidLFielding:
This PR adds implementations for lock/unlock/release/unrelease requests to file management service.

Tests are also added for all of these requests. I also updated logging requests to use the recommended % with argument form.

I will remove verbose prints of request responses over time. While I'm developing I like to see the responses as I'm tweaking them.

Next PR will focus on content requests unless there are specific requests that other developers need sooner. 

At some point we need to record (log) username of client making requests. I was going to add this (since we now have session info from updated auth) here but arrived at a clean cutoff point and decided to do PR. I'll try to work this into upcoming PRs.

I will update description if I think of anything else. Onward...

### DIFF
--- a/filemanager/routes/upload_api.py
+++ b/filemanager/routes/upload_api.py
@@ -19,6 +19,9 @@ from filemanager.controllers import upload
 
 blueprint = Blueprint('upload_api', __name__, url_prefix='/filemanager/api')
 
+user_id = ''
+scope = ''
+username = ''
 
 def is_owner(session: auth_domain.Session, upload_id: str, **kwargs) -> bool:
     """User must be the upload owner, or an admin."""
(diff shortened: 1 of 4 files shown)


filemanager/routes/upload_api.py
```diff
@@ -19,6 +19,9 @@
 
 blueprint = Blueprint('upload_api', __name__, url_prefix='/filemanager/api')
 
+user_id = ''
+scope = ''
+username = ''
```
erickpeirson: Did you mean to leave these here?

filemanager/routes/upload_api.py
```diff
@@ -27,9 +30,12 @@ def is_owner(session: auth_domain.Session, upload_id: str, **kwargs) -> bool:
     upload_obj = uploads.retrieve(upload_id)
     if upload_obj is None:
         return True
+    global user_id, username, scope
```
erickpeirson: This may cause unexpected results. If you need something global, [``flask.g``](http://flask.pocoo.org/docs/1.0/appcontext/#storing-data) is your friend. But any reason not to just rely on the ``request`` object (which is request-global) if you need to access these elsewhere?

PR #15, opened 2018-08-21
Arxivng 1098 1099

DavidLFielding:
Branch ARXIVNG-1098-1099 implements fixes described in tasks ARXIVNG-1098 and ARXIVNG-1099.

ARXIVNG-1098 ensures that file list is updated and current. Additional tests added to demonstrate deleted file is correctly removed from subsequence upload summaries.

ARXIVNG-1099 adds the modified_datetime file property to the list of uploaded files.

These changes should improve integration with UI.

I will remove verbose test message in the future. I also plan to perform additional cleanup work on the file list code in the process/upload.py and controller/upload.py files.

### DIFF
--- a/filemanager/controllers/upload.py
+++ b/filemanager/controllers/upload.py
@@ -565,6 +565,24 @@ def upload_summary(upload_id: int) -> Response:
             raise NotFound(UPLOAD_NOT_FOUND)
         else:
             logger.info("%s: Upload summary request.", upload_db_data.upload_id)
+
+            # Create Upload object
+            upload_workspace = filemanager.process.upload.Upload(upload_id)
+            file_list = upload_workspace.create_file_list()
+
+            details_list = []
+            for fileObj in file_list:
+                file_details = {
+                    'name': fileObj.name,
+                    'public_filepath': fileObj.public_filepath,
+                    'size': fileObj.size,
+                    'type': fileObj.type_string,
+                    'modified_datetime': fileObj.modified_datetime
+                }
+                if not fileObj.removed:
+                    details_list.append(file_details)
+
+
             status_code = status.HTTP_200_OK
             response_data = {
                 'upload_id': upload_db_data.upload_id,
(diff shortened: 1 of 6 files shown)


filemanager/controllers/upload.py
```diff
@@ -565,15 +565,33 @@ def upload_summary(upload_id: int) -> Response:
...
             response_data = {
                 'upload_id': upload_db_data.upload_id,
                 'created_datetime': upload_db_data.created_datetime,
                 'modified_datetime': upload_db_data.modified_datetime,
                 'start_datetime': upload_db_data.lastupload_start_datetime,
                 'completion_datetime': upload_db_data.lastupload_completion_datetime,
-                'files': json.loads(upload_db_data.lastupload_file_summary),
-                'errors': json.loads(upload_db_data.lastupload_logs),
+                'files': details_list,
+                'errors': [],
```
erickpeirson: Do you mean to not return any errors?

DavidLFielding: Technically there are no errors generated from the upload summary request. If we return the errors from the previous upload request there will most likely be 'stale' error messages like "Removed hidden file" or "Changed file name" that the user has already seen and that no longer apply to the current operation. Is this preferable? I expect persistent errors, something like the oversize error, will eventually be included here (once oversize is implemented). At one point I was planning to include error messages with the file data but we determined having a single set of error messages that include file names would allow the UI to present these with the files (while eliminating duplication). Another option, to display current errors, would be to reprocess upload files. For using error messages from previous requests the tricky piece would be to determine which old error messages are still relevant to client. I could scan previous upload errors and include only those where the file still exists in the upload workspace. This would show the most recent action on the file. On the other hand reprocessing would not be too expensive.

Do we want to store errors from previous requests for long term presentation? The upload summary request doesn't do anything so would not generate any errors on its own. The persistent error I might return, like oversize, is not implemented yet. 

So to make a long answer short I did not intend to return any errors at this time.

My plan is to return any errors that make the state of the upload 'ERRORS'.
 [possible states are "READY", "READY_WITH_WARNINGS",  "ERRORS"] 

Maybe 'ERRORS' should be "HAS_ERRORS" or "HAS_FATAL_ERRORS" instead.

We can discuss this more if you like.

PR #17, opened 2018-08-28
ARXIVNG-1131 modified time is updated on unpacked files; also fixed mtime localization

erickpeirson:
The issue was that unpacked files retain their modified timestamp from when they were originally created. Using ``os.utime`` to update the access/modified timestamps to now. 

Also switched to ``datetime.utcfromtimestamp`` to get the correct UTC datetime from the modified time.

### DIFF
--- a/filemanager/arxiv/file.py
+++ b/filemanager/arxiv/file.py
@@ -6,7 +6,11 @@ import os.path
 import re
 from datetime import datetime
 
+from arxiv.base import logging
+
 from filemanager.arxiv.file_type import guess, _is_tex_type, name
+
+logger = logging.getLogger(__name__)
 
 
 class File:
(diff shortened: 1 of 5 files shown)


filemanager/arxiv/file.py
```diff
@@ -59,15 +63,19 @@ def base_dir(self, base: str) -> None:
...
         pdir = os.path.dirname(self.filepath)
         if pdir == self.__base_dir:
             return ''
         elif self.dir:
             return pdir.replace(self.base_dir + '/', "")
         else:
-            # For directories self.dir is empty, must get rest of path from filepath
+            # For directories self.dir is empty, must get rest of path from
+            # filepath
             public_dir = pdir.replace(self.base_dir + '/', "")
```
JaimieMurdock: The `.replace` trick works, but for future reference:

```
os.path.relpath(pdir, start=self.base_dir)
```

erickpeirson: Thanks; I think that's something from @DavidLFielding

PR #18, opened 2018-08-28
ARXIVNG-1136 Fix for security vulnerability that allows files to escape working directory

erickpeirson:
Added a check during extraction to ensure that the destination of the extracted file falls within the upload source directory.


filemanager/utilities/unpack.py
```diff
@@ -113,12 +113,23 @@ def unpack_archive(upload: Upload) -> None:
                             # These get handled in checks and logged.
 
                             # Extract files and directories for now
+                            dest = os.path.join(target_directory, tarinfo.name)
+                            # Tarfiles may contain relative paths! We must
+                            # ensure that each file is not going to escape the
+                            # upload source directory _before_ we extract it.
+                            if source_directory not in os.path.normpath(dest):
+                                continue
```
JaimieMurdock: The semantics here are still confusing - I feel like I'd have to rerun this mentally on reading it again. Proposed alternative:

```
norm_dest = os.path.normpath(dest)
if not norm_dest.startswith(source_directory):
```

One possible bug this avoids: 
```
source_directory = '/my_source_dir'
target_directory = '/my_target_dir'
dest = '../my_source_dir'
assert source_directory in os.path.normpath(dest)
assert target_directory not in os.path.normpath(dest)
```

Also is this really supposed to be `source_directory not in ...` and not `target_directory not in ...` current behavior skips anything that wouldn't be extracted to `source_directory`. I think you want to skip anything that wouldn't be extracted to `target_directory`.

erickpeirson: `startswith` is a good idea. `target_directory` may be a subdirectory arbitrary levels below `source_directory`; I guess it didn't seem consequential for the tarfile to have relative paths, just so long as it doesn't escape `source_directory`

PR #21, opened 2018-09-18
Arxivng 852 1150

DavidLFielding:
This PR implements ARXIVNG-852 to clean out a few extraneous directories from upload and ARXIVNG-1150 that implements downloading file management service logs and upload workspace source logs. 

It's not clear whenever one of the directories that is supposed to get cleaned up in legacy system will 'exist' in arXivNG. The 'processed' directory appears to occur when source files are copied from prior version of paper. This replacement functionality has not been investigated yet and may be implemented differently in new submission system.

Rearranged content download test to come before the workspace is deleted. Raises the question of whether we should add exceptions when situations like this occur.

A few major external events have impacted progress on this PR. Most of these issues have been resolved. There are two API requests that still need to be implemented (download individual files, replace entire content w/backup) and additional upload checks for analyzing submission size (providing total size of upload and oversize calculations). I'm currently thinking that the size calculation would contribute the most to the UI side of things, followed by the individual file download, and then content replacement.


DavidLFielding:
Test file would help, wouldn't they!! Added and pushed to branch.

PR #23, opened 2018-09-21
ARXIVNG 1195

DavidLFielding:
This PR covers ARXIVNG-1195 and implements individual file 'exists' and 'download' requests. This is basically HEAD/GET <file> content with a relative file argument. This is one of the last critical requests needed for a fully functional submission UI interface.

I have also added the size and modify date time to all HEAD 'exists' requests for content/individual files/logs. This uncovered a bug in content head routine.

If the content download package does not already exist the checksum is always null.  If content package doesn't exists the HEAD request never builds the package. This means if you followed protocol and do a head request first you will NEVER get the OK (checksum) to download the content package.

I added a call in content existence check routine to build content package when the content package file does not exist.

I still need to implement content replace request for admins.

Nice to end the week by wrapping up open task with tests, bug fixes,  and a PR.

### DIFF
--- a/filemanager/controllers/upload.py
+++ b/filemanager/controllers/upload.py
@@ -31,6 +31,7 @@ logging.basicConfig(level=logging.DEBUG)
 
 logger = logging.getLogger(__name__)
 
+
 def _get_service_logs_directory() -> str:
     """
     Return path to service logs directory. We may eventually
@@ -43,6 +44,7 @@ def _get_service_logs_directory() -> str:
     """
     config = get_application_config()
     return config.get('UPLOAD_SERVICE_LOG_DIRECTORY', '')
+
 
 service_log_path = os.path.join(_get_service_logs_directory(), 'upload.log')
 
@@ -524,7 +526,7 @@ def upload(upload_id: int, file: FileStorage, archive: str,
 
             response_data = {
                 'upload_id': upload_db_data.upload_id,
-                'upload_total_size':upload_workspace.total_upload_size,
+                'upload_total_size': upload_workspace.total_upload_size,
                 'created_datetime': upload_db_data.created_datetime,
                 'modified_datetime': upload_db_data.modified_datetime,
                 'start_datetime': upload_db_data.lastupload_start_datetime,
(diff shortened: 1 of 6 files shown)


filemanager/controllers/upload.py
```diff
@@ -931,6 +958,148 @@ def get_upload_content(upload_id: int) -> Response:
...
+        upload_workspace = filemanager.process.upload.Upload(upload_id)
+
+        # Returns path if file exists
+        path_to_file = upload_workspace.get_content_file_path(public_file_path)
+
+        checksum = ''
+
+
+        if path_to_file:
+            size = os.path.getsize(path_to_file)
```
erickpeirson: Would it make sense to provide this as a method on the workspace, so that the filesystem is more cleanly screened off from the controller?

filemanager/controllers/upload.py
```diff
@@ -907,9 +909,34 @@ def check_upload_content_exists(upload_id: int) -> Response:
...
+    # If content package doesn't exist it's difficult to calculate
+    # the checksum or for it ever to show it exists via the HEAD
+    # request.
+    if not os.path.exists(content_path):
+        # At least try to build content package, we don't do much
+        # error checking at this point.
+        upload_workspace.pack_content()
+
+    # Let's return size and last modification date
+    if os.path.exists(content_path):
```
erickpeirson: These ``os.path.exists`` calls might also be good candidates for offering as a method on the workspace

filemanager/controllers/upload.py
DavidLFielding: Absolutely. I was thinking the same thing as I was wrapping up the PR. The workspace is in a way leaking functionality that it should implement itself. This cleanup will involve multiple routines (that should probably rely on underlying File object) but I will definitely try to clean this up asap.

filemanager/controllers/upload.py
DavidLFielding: I left this in there as a test just to see of you'd catch it. :^) I will fix it.

erickpeirson:
This looks great to merge!

PR #27, opened 2018-12-14
ARXIVNG-484

DavidLFielding:
This PR focuses on end of file cleanup and fixing carriage returns in files generated on PC and MAC platforms.

This cleanup is necessary because compilation fails when files are not cleaned up properly.

Jake indicates that unmacify running in production fails in certain cases. The Python implementation is new and improved from Perl implementation and may not experience similar failures. I will test this once I receive some test files from Jake.


filemanager/process/upload.py
```diff
@@ -999,10 +1015,140 @@ def _add_file(fpath: str, warnings: list, errors: list) -> File:
...
-        self.log(f'Unmacify file {file_name}')
-        self.log(f"I'm sorry Dave I'm afraid I can't do that. unmacify not implemented YET.")
+        self.log(f'Un{type}ify file {file_obj.filepath}')
+
+        # Open file and look for carriage return.
+        #
+        new_filepath = filepath + ".new"
+        with open(filepath, 'rb', 0) as infile, \
+            open(new_filepath, 'wb', 0) as outfile, \
+                mmap.mmap(infile.fileno(), 0, access=mmap.ACCESS_READ) as s:
```
JaimieMurdock: Would you mind commenting a little on the decision to use `mmap.mmap` instead of `open` for the file `s` or why `s` is necessary given that its pointing to the same pointer as `infile`? Confused by the conventions, suspect I have something to learn having not used `mmap` before.

filemanager/process/upload.py
```diff
@@ -72,6 +75,19 @@ def __init__(self, upload_id: int):
         # not upload or delete files. Those requests update total size.
         self.calculate_client_upload_size()
 
+    # Debug
+    def set_debug(self, set: bool):
+        """
+        Activate/deactivate debugging
+        :param set:
```
erickpeirson: Don't forget to use numpy style docstrings, e.g. parameters are formatted differently

filemanager/process/upload.py
```diff
@@ -999,10 +1015,140 @@ def _add_file(fpath: str, warnings: list, errors: list) -> File:
...
+        # Check for special characters at end of file.
+        # Remove EOT/EOF
+
+        # Get the absolute file path
+        filepath = file_obj.filepath
+
+        if self.debug():
+            self.log(f"Checking file termination for {filepath}.")
+
+        if os.path.exists(filepath):
```
erickpeirson: One way to cut down on indentation here would be to return early. Ie 

```
if not os.path.exists(filepath):
    return
```

We'll want to aim for 79-character line limits as we get further along

filemanager/process/upload.py
```diff
@@ -999,10 +1015,140 @@ def _add_file(fpath: str, warnings: list, errors: list) -> File:
...
+        and line feeds.
+
+        Jake informs me there is a bug in the Perl unmacify routine.
+
+        :param file_name:
+        :return:
+        """
+
+        # Determine type of file we are dealing with PC or MAC
+        type = 'mac'
```
erickpeirson: ``type`` is a built-in; should probably use a different name

filemanager/process/upload.py
```diff
@@ -999,10 +1015,140 @@ def _add_file(fpath: str, warnings: list, errors: list) -> File:
...
+        self.log(f'Un{type}ify file {file_obj.filepath}')
+
+        # Open file and look for carriage return.
+        #
+        new_filepath = filepath + ".new"
+        with open(filepath, 'rb', 0) as infile, \
+            open(new_filepath, 'wb', 0) as outfile, \
+                mmap.mmap(infile.fileno(), 0, access=mmap.ACCESS_READ) as s:
+
+            if type == 'pc':
```
erickpeirson: It might be a good idea to define these values as globals in the module. Advantages are that it reduces the task of undetected typos, and is easier to see what the possible values are. E.g.

```python
PC = 'pc'
MAC = 'mac'

...

if file_type == PC:
    ...
```

filemanager/process/upload.py
DavidLFielding: I fell down on my documentation. I made a pass at improving docstrings in PR #28. I still have a few more files to cleanup. I will continue to improve on docstrings in file manager code.

filemanager/process/upload.py
DavidLFielding: Fixed. I'll keep returning early in mind. I'll also work with pylint to eliminate other long lines.

filemanager/process/upload.py
DavidLFielding: The legacy code reads a file in blocks and I was looking for ways to eliminate some of the crufty code. I came about this solution for executing regexes over large files without reading file in chunks and this solution seems like an elegant way to do this. I will take another look at those references and try to follow up with a better explanation.

DavidLFielding:
I've addressed the feedback above. There is a significant amount of doctoring enhancements in PR#28. Let me know if there are more changes need for this PR.

erickpeirson:
Thanks, looks great!

PR #29, opened 2019-01-20
ARXIVNG-1412 Identify and return submission source format.

DavidLFielding:
The main focus of this PR is task ARXIVNG-1412 which will identify and return submission source format. The source format is now returned in the upload and summary JSON responses.

One **design issue** has come up. The code that identifies the format verifies a Postscript file is valid by calling the Ghostscript GS command. The dilemma I have is whether to perform a simple sys.call or incorporate the entire Celery asynchronous thread management mechanism. I suspect Celery is the preferred way to go. Seems like a lot of overhead for such a narrow check. Will discuss with team. Maybe this could be migrated to compilation service, where a submission containing a single Postscript file is verified and then converted to PDF. On the other hand this would delay correcting the problem.

The determine_source_format() routine is the main routine that attempts to figure out the submission type. Various routines assist this effort by counting the types of files and making sure type suffixes are normalized to use an extension expected by our system.

A significant number of tests have been written to exercise the submission type code along with tests for each of the underlying support routines.

Improvement have been made to several files as recommended by pylint, pydocstyle, and mypy in an ongoing attempt to clean up the file manager code. 

*Bug: I fixed a bug left over from the previous unmacify PR. I may need to merge this sooner than later.

*Bug: Fixed a bug with returning goofy LATEX2e human readable type string.

*Bug: Fixed a problem with mixed int/str type upload_id. I believe we eventually need to convert upload_id to string type.

*Type Question: 'upload_id' is currently type 'int'. I'd like to convert this to type 'str'. Might allow for more flexibility in upload identifiers.

This PR fixes a few problems with previous PRs.


filemanager/process/upload.py
```diff
@@ -1862,8 +1873,341 @@ def source_log_checksum(self) -> str:
...
+                    # TODO: possible to continue.
+                    self.source_format = 'postscript'
+                else:
+                    # TODO: What to do here? 'None' indicates error. This is
+                    # TODO: internal error. Error has been registered. Not
+                    # TODO: much user can do.
+                    pass
+            elif file_type == 'pdf':
+                # Rename
+                new_file_obj = self.fix_file_ext(obj, 'pdf')
```
erickpeirson: It seems odd that we'd be changing files in a method called "determine_source_format". Once we get v0.1 out, for v0.2 I'd be interested in seeing if we can refactor a bit to separate descriptive procedures from fixes. Thinking about long-term maintenance/extension. But OK for now, let's focus on getting the functionality we want.

filemanager/process/upload.py
```diff
@@ -1862,8 +1873,341 @@ def source_log_checksum(self) -> str:
...
+            self.add_error("", msg)
+        else:
+            msg = "No file found."
+            self.add_error("", msg)
+
+        return None
+
+
+    @property
+    def source_format(self) -> str:
```
erickpeirson: Why use getters/setters and a private attribute? Why not just use ``.source_format`` directly?

filemanager/process/upload.py
```diff
@@ -1862,8 +1873,341 @@ def source_log_checksum(self) -> str:
...
+        # Analyze files formats in user upload workspace
+        formats = self.count_file_types()
+
+        if formats['files'] == formats['invalid']:
+            # Were all files have been identified as type 'invalid'
+            msg = ("All files are auto-ignore. If you intended to withdraw the"
+                   " article, please use the 'withdraw' function from the list"
+                   "of articles on your account page.")
+            self.add_warning("", msg)
+            self.source_format = 'invalid'
```
erickpeirson: Why are we setting this? If the public method ``determine_source_format()`` is the preferred way for a caller (e.g. a controller) to get the source format, then why also maintain a public attribute? Some alternatives:

If this method should only run once in a request context, consider:

```python
class Upload:
    def __init__(self, ...):
        self.source_format = self._determine_source_format()

    def _determine_source_format():
        ...
```

If this can run several times in a request context, consider:

```python
class Upload:
    @property
    def source_format(self):   # Rename determine_source_format -> source_format
        ...
```

erickpeirson:
> *Type Question: 'upload_id' is currently type 'int'. I'd like to convert this to type 'str'. Might allow for more flexibility in upload identifiers.

This seems reasonable to me. I think you can just do it.

erickpeirson:
> One **design issue** has come up. The code that identifies the format verifies a Postscript file is valid by calling the Ghostscript GS command. The dilemma I have is whether to perform a simple sys.call or incorporate the entire Celery asynchronous thread management mechanism. I suspect Celery is the preferred way to go. Seems like a lot of overhead for such a narrow check. Will discuss with team. Maybe this could be migrated to compilation service, where a submission containing a single Postscript file is verified and then converted to PDF. On the other hand this would delay correcting the problem.

Per comments in code: I'm really leaning toward moving this to the compilation service.

filemanager/arxiv/file_type.py
```diff
@@ -31,7 +32,7 @@
 
 
 def set_priority() -> int:
```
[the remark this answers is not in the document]
DavidLFielding: Removed.

filemanager/arxiv/file_type.py
```diff
@@ -469,37 +480,46 @@ def is_tex_type(type: str) -> bool:
...
+    """
     if type in type_name.keys():
         return type_name[type]
-    else:
-        return 'unknown'
+
+    return 'unknown'
 
 
 def get_type_priority(type: str) -> int:
```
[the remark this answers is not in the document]
DavidLFielding: I renamed this elsewhere bit missed this occurrence. Done.

filemanager/arxiv/file_type.py
```diff
@@ -469,37 +480,46 @@ def is_tex_type(type: str) -> bool:
...
-    """Return a cleaned up version of the internal file type minus
-    TYPE prefix and lower cased."""
-    (type, tex_format, error) = guess_file_type(filepath)
+    """
+    Return a cleaned up version of the internal file type.
+
+    Removes TYPE_ prefix and lower cases resulting type.
+    """
+    # Not using tex_format or error from guess_file_type at this time
+    (type, _, _) = guess_file_type(filepath)
```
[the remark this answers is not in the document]
DavidLFielding: Done

filemanager/controllers/upload.py
```diff
@@ -1261,24 +1265,24 @@ def __last_modified(filepath: str) -> str:
...
+    """
+    if os.path.exists(service_log_path):
+        return open(service_log_path, 'rb')
 
-        """
-        if os.path.exists(service_log_path):
-            return open(service_log_path, 'rb')
-        else:
-            return ""
+    return ""
```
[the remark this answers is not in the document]
DavidLFielding: Eliminated check and null return so return type is consistent.

filemanager/process/upload.py
```diff
@@ -1862,8 +1873,341 @@ def source_log_checksum(self) -> str:
...
+                self.source_format = 'invalid'
+                self.add_error(public_file_path, msg)
+            elif file_type == 'postscript':
+                # The logic for Postscript format actually verifies that the format
+                # is valid Postscript by running Ghostscript
+                # Rename
+                new_file_obj = self.fix_file_ext(obj, 'ps')
+                if new_file_obj is not None:
+                    obj = new_file_obj
+                    # TODO: NEED TO PERFORM LIVE PS FORMAT CHECK
```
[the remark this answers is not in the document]
DavidLFielding: The compiler service converter Docker image is already set up with Ghostscript so it would be easier to do there.

filemanager/process/upload.py
```diff
@@ -1862,8 +1873,341 @@ def source_log_checksum(self) -> str:
...
+                    # TODO: Do we want to do something else when Postscript
+                    # TODO: file fails to validate? We currently generate
+                    # TODO: warning AND set format to 'ps' (as if it's
+                    # TODO: possible to continue.
```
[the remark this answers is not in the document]
DavidLFielding: One could argue validating formats makes sense in the file upload arena but then we might want to check everything that gets uploaded. I suspect problems with 'ps' files resulted in this test getting added. This reminds me that Jake sometimes checks/fixes images uploaded as part of submissions. I will check with admins on what checks/fixes admins perform that might make sense to eventually automate.

filemanager/process/upload.py
DavidLFielding: I believe this could be moved to main check method. I was thinking the same thing as I was reimplementing source format but wanted to get the initial code working before refactoring. I was worried that if I fiddle too much I might break things and have issues debugging. That said, I suspect these extension fixes don't happen very often. I should have added a note to code about refactoring. How about I add a reminder in code and do this during a subsequent task.

filemanager/process/upload.py
DavidLFielding: I should have relied on source_format instead of determine_source_format and called determine_source_format from within source_format. This might get run multiple times if Upload object remains while files are deleted and added.

Let me change this to access via source_format instead of direct call to determine_source_format.

filemanager/process/upload.py
DavidLFielding: I'm making a change (below) that should resolve this issue.

filemanager/process/upload.py
DavidLFielding: Cleaned out a bunch of cruft. Should look much better now.

erickpeirson:
@DavidLFielding I resolved some conflicts because they were my fault

filemanager/process/upload.py
erickpeirson: Excellent

erickpeirson:
@DavidLFielding  good to go!

PR #32, opened 2019-02-12
ARXIVNG-1440 changes to support upload, compilation workflow

erickpeirson:
In order to keep the submission consistent, it's nice to get status information back after deletion actions. So this change basically returns ``200`` after deletion along with the workspace status, instead of ``204``. Everything else is just tweaks for clarity, minor consolidation.

I am still getting one failing test, and I can't figure out why it would be related to these changes:

```
======================================================================
FAIL: test_check_file_unmacify (tests.test_process_upload.TestInternalSupportRoutines)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/brp53/projects/arxiv-filemanager/tests/test_process_upload.py", line 325, in test_check_file_unmacify
    self.assertTrue(is_same, 'Eliminated unwanted CR characters from MAC file.')
AssertionError: False is not true : Eliminated unwanted CR characters from MAC file.

----------------------------------------------------------------------
```

### DIFF
--- a/filemanager/controllers/upload.py
+++ b/filemanager/controllers/upload.py
@@ -28,7 +28,7 @@ from filemanager.shared import url_for
 
 from filemanager.domain import Upload
 from filemanager.services import uploads
-
+from filemanager.process.upload import Upload as UploadWorkspace
 from filemanager.arxiv.file import File
 
 # Temporary logging at service level - just to get something in place to build on
@@ -177,7 +177,7 @@ def delete_workspace(upload_id: int) -> Response:
                             "current state is '%s'", upload_id, upload_db_data.state)
                 raise NotFound(UPLOAD_WORKSPACE_NOT_FOUND)
 
-            upload_workspace = filemanager.process.upload.Upload(upload_id)
+            upload_workspace = UploadWorkspace(upload_id)
 
             # Call routine that will do the actual work
             upload_workspace.remove_workspace()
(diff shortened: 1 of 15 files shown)


filemanager/controllers/upload.py
```diff
@@ -28,7 +28,7 @@
 
 from filemanager.domain import Upload
 from filemanager.services import uploads
-
+from filemanager.process.upload import Upload as UploadWorkspace
```
JaimieMurdock: ohhh. Process module. Nice!

JaimieMurdock:
@DavidLFielding will have to provide guidance on the Mac CRLF issues.

DavidLFielding:
Looks good. 

I checked out fresh copy of 1440 and tests run with no errors. I've seen leftover crud from previous runs interfere with subsequent tests and have recently added remove_workspace call to cleanup work area after each test completes.

erickpeirson:
That's probably what's going on. Will hook up Travis, too, so we get clean builds.

erickpeirson:
So this is interesting. Sometimes on Travis this also fails, but sometimes not.

```
======================================================================
FAIL: test_check_file_unmacify (tests.test_process_upload.TestInternalSupportRoutines)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/arXiv/arxiv-filemanager/tests/test_process_upload.py", line 325, in test_check_file_unmacify
    self.assertTrue(is_same, 'Eliminated unwanted CR characters from MAC file.')
AssertionError: False is not true : Eliminated unwanted CR characters from MAC file.
----------------------------------------------------------------------
```

erickpeirson:
I wonder if somehow the ``stat`` of the file is sometimes not obviously changing, hence [``filecmp.cmp``](https://docs.python.org/3/library/filecmp.html#filecmp.cmp) returns ``False``. Maybe if it's super fast. Will try setting ``shallow=False``, and we can monitor on future builds.

erickpeirson:
Well, that wasn't it.

erickpeirson:
@DavidLFielding Ok, got it. I missed the use of ``filecmp.cmp`` in the ``unmacify()`` method itself. Doing a deep comparison there fixed the test. In the meantime, I slightly rewrote that test to actually look for carriage returns in the content of the file. Will wait to merge until you have a chance to verify that I didn't muck this up.

DavidLFielding:
I didn't realize a filecmp.cmp equal comparison result against a file that contains no carriage returns is not sufficient. The new explicit check looks good.

erickpeirson:
It probably is sufficient; I was just being paranoid

PR #33, opened 2019-02-18
ARXIVNG-485

DavidLFielding:
This PR implements ARXIVNG-485: Upload check: Check Postscript file (edits postscript).

This work is limited to code that removes embedded Thumbnail/Postscript/Preview images from Postscript files. The legacy routines include/call other functions (embedded fonts/TIFF bitmap removal) that will be handled in separate tickets due to the amount of work involved.

These embedded formats result in failures later down the submission workflow. I suspect these failures occur when we attempt to convert from Postscript to PDF and possibly even with compilation.

Tests compare test file we just stripped to a reference copy of a valid stripped file.

Tests are broken due to changes that were merged with develop (not expecting checksum I result?). I must have pulled develop before the tests were fixed. Not sure if I should try to fix in my feature branch.

Run the tests that check Postscript files and strip out any unwanted embedded formats.

    nose2 -v tests.test_process_upload.TestInternalSupportRoutines.test_check_postscript

Also contains a fix for dual ps/postscript source format. Using 'ps'.


DavidLFielding:
Merged 1440 to fix problem with tests.

filemanager/process/upload.py
```diff
@@ -1301,11 +1322,244 @@ def unmacify(self, file_obj: File):
...
+
+        Returns
+        -------
+            Need to decide if we need to return anything.
+
+        """
+        self.add_warning(file_obj, "strip_tiff() NOT IMPLEMENTED")
+
+
+    def strip_preview(self, file_obj: File, what_to_strip: str):
```
erickpeirson: minor: missing return annotation ``-> None:``

filemanager/process/upload.py
```diff
@@ -1301,11 +1322,244 @@ def unmacify(self, file_obj: File):
...
+        ----------
+        file_obj : File
+            File to strip embedded preview from.
+        what_to_strip : str
+            The type of inclusion that we are seeking to remove [Thumbnail,
+            Preview, Photoshop]
+
+        Returns
+        -------
+            File preview is removed from file.
```
erickpeirson: minor: not really a return -- move to summary, below line 1350?

filemanager/process/upload.py
```diff
@@ -1301,11 +1322,244 @@ def unmacify(self, file_obj: File):
...
+
+        """
+        if self.debug():
+            self.log(f"Strip embedded '{what_to_strip}' from file '{file_obj.name}'.")
+
+        # Set start and end delimiters of preview.
+        if what_to_strip == 'Photoshop':
```
erickpeirson: I'd suggest using constants or enums, e.g.

```python
PHOTOSHOP = 'Photoshop'
...
def strip_preview(self, file_obj: File, what_to_strip: str):
    if what_to_strip == PHOTOSHOP:
```

filemanager/process/upload.py
```diff
@@ -1301,11 +1322,244 @@ def unmacify(self, file_obj: File):
...
+            end_re = b'^%EndPhotoshop'
+        elif what_to_strip == 'Preview':
+            start_re = b'^%%BeginPreview'
+            end_re = b'^%%EndPreview'
+        elif what_to_strip == 'Thumbnail':
+            start_re = b'Thumbnail'
+            end_re = b'^%%EndData'
+
+        # Open a file to store stripped contents
+        dir = file_obj.base_dir
```
erickpeirson: ``dir()`` is a built-in; probably want a different name for this

filemanager/process/upload.py
```diff
@@ -1301,11 +1322,244 @@ def unmacify(self, file_obj: File):
...
+                    strip_warning = strip_warning + f" to line {line_no},"
+                    retain = True
+                    # Handle bug in certain files
+                    # AI bug %%EndData^M%%EndComments
+                    if re.search(b'.*\r%/%', line):
+                        outfile.write(line)
+
+                line_no = line_no + 1
+
+            infile.close()
```
erickpeirson: Since you're using ``open`` as a context manager (lines 1389, 1390), you shouldn't need to explicitly ``.close()`` this pointers; ``.close()`` is called when the context (block) exits

filemanager/process/upload.py
```diff
@@ -1301,11 +1322,244 @@ def unmacify(self, file_obj: File):
...
+            outfile.close()
+
+            # Generate some warnings
```
erickpeirson: Consider outdenting lines 1424-1440 so that this comes outside the ``with open(...`` block

filemanager/process/upload.py
DavidLFielding: Fixed

filemanager/process/upload.py
DavidLFielding: Moved.

filemanager/process/upload.py
DavidLFielding: Fixed here and in calling routine.

filemanager/process/upload.py
DavidLFielding: Renamed.

filemanager/process/upload.py
DavidLFielding: I'm a little confused on this one since lines 1424-1440 rely on variables that will only get set within the "with open" scope. If "with open" fails this code doesn't stand on its own and I don't want it to run. That said, I suspect I can outdent it and make it work. I'll take a closer look.

erickpeirson: But if ``with open`` fails, this won't run in any case because an exception will be thrown

PR #34, opened 2019-02-25
ARXIVNG-491

DavidLFielding:
ARXIVNG-491: Upload check: Strip non-compliant embedded TIFF bitmaps from Postscript file

This code is similar to previous PR that strips preview/thumbnails/photoshop from Postscript files. These TIFF bitmaps must be remove to avoid problems downstream in the compilation/conversion workflow.

This is a small routine that remove a trailing embedded TIFF bitmap tacked on to a Postscript file. 

Note: There is other code that removes leading TIFF bitmaps in the repair_dos_eps() that still needs to be implemented.


filemanager/process/upload.py
```diff
@@ -1301,11 +1322,300 @@ def unmacify(self, file_obj: File):
...
+                    offset = len(line)
+                    end = infile.tell() - offset
+                    infile.seek(end,1)
+
+                    msg = f"No %%EOF, but truncate at {end} bytes, " \
+                          f"lastnonwhitespace was {lastnw}  untruncated " \
+                          f"version moved to $scratch_file"
+                    self.log(msg)
+
+                if re.search(b'\S', line):
```
erickpeirson: Lost me here; what's this for?

filemanager/process/upload.py
```diff
@@ -1301,11 +1322,300 @@ def unmacify(self, file_obj: File):
...
+                    lastnw = line
+
+
+            # Truncate file after EOF marker
+            if end:
+                infile.truncate(end)
+                infile.close()
```
erickpeirson: Don't need this close(), since you're in a ``with open(..`` block

consider outdenting lines 1402 and 1403 so that they are outside the block

filemanager/process/upload.py
```diff
@@ -1301,11 +1322,300 @@ def unmacify(self, file_obj: File):
...
+            File with TIFF to be stripped.
+
+        Returns
+        -------
+            Need to decide if we need to return anything.
+
+        """
+        self.log(f"checking '{file_obj.public_filepath}' for TIFF")
+
+        filepath = file_obj.filepath
```
erickpeirson: Minor nit: any particular reason not to just use ``file_obj.filepath`` directly? E.g. (L1360) ``with open(file_obj.filepath, 'rb+', 0) as infile:``

DavidLFielding: I believe there are times when line length comes in to play and I try to shorten my verbose variables and methods. I will take a look.

erickpeirson: Ok, fair enough; I do that sometimes too

filemanager/process/upload.py
DavidLFielding: In the exception case, where the proper Postscript %%EOF is not detected, we log the last line with stuff on it before the TIFF image. The TIFF is removed. I'll add a comment.

filemanager/process/upload.py
DavidLFielding: Removed close. Lines 1402-1403 are only relevant in block  "if end:" where TIFF image is truncated. I believe outdenting outside block would mean "TIFF removed" log message is generated regardless of whether TIFF is actually removed.

PR #36, opened 2019-03-14
updated dependencies, minor bugfix

erickpeirson:
There was a weird bug in HEAD requests to ``check_upload_content_exists()``. Basically the ``Content-Length`` header was getting set twice: we were setting ``Content-Length`` correctly based on the size of the resource, but then Flask was also setting the header based on the size of the response. So this fix makes sure that we are only returning our (correct) header.

### DIFF
--- a/filemanager/routes/upload_api.py
+++ b/filemanager/routes/upload_api.py
@@ -223,22 +223,21 @@ def unrelease(upload_id: int) -> tuple:
 
 @blueprint.route('/<int:upload_id>/content', methods=['HEAD'])
 @scoped(scopes.READ_UPLOAD)
-def check_upload_content_exists(upload_id: int) -> tuple:
+def check_upload_content_exists(upload_id: int) -> Response:
     """
     Verify that upload content exists.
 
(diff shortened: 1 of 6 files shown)


filemanager/routes/upload_api.py
```diff
@@ -229,7 +229,13 @@ def check_upload_content_exists(upload_id: int) -> tuple:
     Returns an ``ETag`` header with the current source package checksum.
     """
     data, status_code, headers = upload.check_upload_content_exists(upload_id)
-    return jsonify(data), status_code, headers
+    response = jsonify(data)
```
erickpeirson: Here's the meat of the PR

erickpeirson:
Actually it probably is; just focused on this one since it was raising issues downstream. Will take a look at the other two before merging

erickpeirson:
@DavidLFielding Ok, added some tests to capture what was going on (553bff5) and then fixed the other endpoints (d1e9d46). Also took the opportunity to update some of the type annotations.

PR #37, opened 2019-03-20
ARXIVNG-1301 support for compressed package size in workspace status response

erickpeirson:
This adds ``upload_compressed_size`` to the upload workspace status response. 

I added a few tests, which surfaced some unexpected behavior around the disposition of the packed content tarball. I moved tests for the ``process.upload`` module into ``process.tests``, so that I could run them separately (e.g. ``pipenv run nose2 -s filemanager/process/``).

Notable changes:
- Blowing away the packed content tarball when some or all files are deleted from the workspace.
- Not packing the workspace when there are no files, and returning ``0`` size when there are no files.

I added a few notes with ideas for refactoring/decomposing, when we're ready to think about that.


DavidLFielding:
"I moved tests for the process.upload module into process.tests, so that I could run them separately (e.g. pipenv run nose2 -s filemanager/process/)."

I run these upload tests separately all the time and often run a single test using something like:

    nose2 -v tests.test_process_upload.TestInternalSupportRoutines.test_strip_tiff
    or
    nose2 -v tests.test_process_upload

so I'm trying to understand the benefit of moving the test file itself. I tend to sit in the tests directory and run the tests I want so maybe I need to change how I test. Is the idea to disperse all tests into the code subdirectories?

erickpeirson:
> I tend to sit in the tests directory and run the tests I want so maybe I need to change how I test. Is the idea to disperse all tests into the code subdirectories?

Good question. When we started out, I also was keeping tests in one big directory. After our team discussion around testing strategies, I started keeping tests closer to the code to which they pertain. But we should be consistent within a project -- I'll move those back on the next PR.

PR #39, opened 2019-04-08
ARXIVNG-2020 wired up readiness endpoint to check dependencies

erickpeirson:
This extends the service status endpoint to function as a readiness endpoint. Returns 200 OK if the database and filesystem are available. Returns 503 Service Unavailable if either or both are unavailable.

This also introduces a Helm chart for deployment filemanager, and an application profile that describes the application dependencies and other important details.

@DavidLFielding Tagging you on this so you have a chance to see what I'm doing to deploy this.


profile.yml
```diff
@@ -0,0 +1,31 @@
+# Application profile
```
erickpeirson: This is something that we'll start doing for all applications that we deploy in k8s

wsgi.py
```diff
@@ -4,9 +4,11 @@
 import os
 
 
+__flask_app__ = create_web_app()
```
erickpeirson: The app needs to stick around between requests; see also https://github.com/arXiv/arxiv-browse/pull/69 and https://github.com/arXiv/arxiv-browse/pull/79

PR #41, opened 2019-05-08
ARXIV-2032

DavidLFielding:
This PR focuses on implementing checkpoint support in the File Manager service (intended client is file upload page in submission UI). 

ARXIVNG-2032: Provide endpoints for admin functions on user submission

Independent API checkpoint requests allow you to create a checkpoint, list_checkpoints, remove_checkpoint, remove_all_checkpoints, restore_checkpoint, and download checkpoint file.

User is responsible for all actions when using individual API requests. For example, if you wanted to manually create a checkpoint and then follow with normal upload request you would need to delete all files unless you want the subsequent upload to include all previous files.

At the current time, 'upload with checkpoint' first creates a checkpoint and then deletes all existing files before unpacking newly uploaded files. It's possible we simply want to add files to the current workspace and not delete previous files. **Do admins want to keep existing files and upload a single file?**  **Do we want the removal of existing files to be an option?**
 
New auth release 0.3.4rc1 contains the checkpoint scopes needed to test this PR. Pipfile has been updated to reflect this.

* I'm still thinking about the 'checkpoint_with_upload' path component name in upload route.

upload with checkpoint:
@blueprint.route('<int:upload_id>/checkpoint_with_upload', methods=['POST'])

where 'checkpoint' exposes independent checkpoint request:
@blueprint.route('<int:upload_id>/checkpoint', methods=['POST'])

One option is to change independent request to 'create_checkpoint' and simplify upload path to 'checkpoint'. Or change "checkpoint_with_upload" to "with_checkpoint" though "checkpoint_with_upload" implies the order in which checkpoint comes before upload.


DavidLFielding:
Fixed test failure (typo in test filename).

I'm trying to figure out why file-level comparison between checkpoint contents and original uploaded gzipped tar are failing on Travis. This tests works fine in my environment but seems to fail in Travis. I'm currently investigating.

filemanager/controllers/upload.py
```diff
@@ -387,12 +395,17 @@ def upload(upload_id: Optional[int], file: FileStorage, archive: str,
     #                              task_id=result.task_id)}
     # return ACCEPTED, status.ACCEPTED, headers
     # End delete
+    username = ''
+    if user:
+        username = user.username;
+    else:
+        username = 'NA'
```
erickpeirson: What's the reason for using ``'NA'`` instead of `None`?

filemanager/controllers/upload.py
```diff
@@ -479,7 +492,12 @@ def upload(upload_id: Optional[int], file: FileStorage, archive: str,
         # Create Upload object
         upload_workspace = UploadWorkspace(upload_id)
 
-        # Process upload_db_data
+        # checkpoint - create backup of user files in workspace
+        if checkpoint:
+            checksum = upload_workspace.create_checkpoint(user)
+            upload_workspace.client_remove_all_files()
```
erickpeirson, line 530: This seems surprising to me given the semantics so far -- does creating a checkpoint automatically entail deleting all files in the active workspace? Seems like that ought to be indicated/controlled separately.

filemanager/controllers/upload.py
```diff
@@ -1319,6 +1337,420 @@ def get_upload_service_log() -> Response:
...
+    Parameters
+    ----------
+    upload_id : int
+        The unique identifier for upload workspace.
+    use : str
+        User making create checkpoint request.
+
+    Returns
+    -------
+    Standard Response tuple containing response content, HTTP status, and HTTP headers.
```
erickpeirson:
```suggestion
    tuple
        Standard Response tuple containing response content, HTTP status, and 
        HTTP headers.
```

filemanager/controllers/upload.py
```diff
@@ -1319,6 +1337,420 @@ def get_upload_service_log() -> Response:
...
+
+    """
+    user_id = str(user.username)
```
erickpeirson: Consider using [``user_id``](https://arxiv.github.io/arxiv-auth/arxiv.users/arxiv.users.domain.html#arxiv.users.domain.User.user_id) instead of the username? The username isn't guaranteed to stay the same.

filemanager/controllers/upload.py
```diff
@@ -1319,6 +1337,420 @@ def get_upload_service_log() -> Response:
...
+    logger.info("%s: Create checkpoint ['%s'].", upload_id, user_id)
+
+    try:
+        # Make sure we have an upload_db_data to work with
+        upload_db_data: Optional[Upload] = uploads.retrieve(upload_id)
```
erickpeirson: This might be slightly out of scope, but while I'm thinking about it: consider raising an exception in ``uploads.retrieve()`` if the upload ID does not exist (rather than returning None).

erickpeirson: So in this controller it would look like:

```python
try:
    upload_db_data: Upload = uploads.retrieve(upload_id)
except uploads.UploadDoesNotExist as e:    # For example 
    raise NotFound(UPLOAD_NOT_FOUND) from e
```

filemanager/controllers/upload.py
```diff
@@ -1319,6 +1337,420 @@ def get_upload_service_log() -> Response:
...
+        # Store in DB
+        #   uploads.update(upload_db_data)
+        ###
+
+        response_data = {'reason': UPLOAD_CREATED_CHECKPOINT}  # Get rid of pylint error
+        status_code = status.OK
+
+    except IOError:
+        logger.error("%s: Lock workspace request failed ", upload_id)
+        raise InternalServerError(CANT_DELETE_FILE)
```
erickpeirson, line 1471: Is this the right error message?

filemanager/controllers/upload.py
```diff
@@ -1319,6 +1337,420 @@ def get_upload_service_log() -> Response:
...
+        #    logger.info("%s: Lock: Workspace is already locked.", upload_id)
+        #else:
+        #    upload_db_data.lock = Upload.LOCKED
+        #
+        # Store in DB
+        #   uploads.update(upload_db_data)
+        ###
+
+        response_data = {'reason': UPLOAD_CREATED_CHECKPOINT}  # Get rid of pylint error
+        status_code = status.OK
```
erickpeirson, line 1467: Minor nit: looks like if we get all the way to here, the only possible path is L1400 -- so no need for a new variable

filemanager/controllers/upload.py
```diff
@@ -1319,6 +1337,420 @@ def get_upload_service_log() -> Response:
...
+    Parameters
+    ----------
+    upload_id : int
+        The unique identifier for upload workspace.
+    user : str
+        User making create checkpoint request.
+
+    Returns
+    -------
+    Standard Response tuple containing response content, HTTP status, and HTTP headers.
```
erickpeirson:
```suggestion
    tuple
        Standard Response tuple containing response content, HTTP status, and 
        HTTP headers.
```

filemanager/controllers/upload.py
```diff
@@ -1319,6 +1337,420 @@ def get_upload_service_log() -> Response:
...
+
+        Response content includes details for each checkpoint.
+
+    """
+    user_id = str(user.username)
```
erickpeirson: Same here -- recommend using user_id

filemanager/controllers/upload.py
```diff
@@ -1319,6 +1337,420 @@ def get_upload_service_log() -> Response:
...
+        if upload_db_data is None:
+            # Invalid workspace identifier
+            raise NotFound(UPLOAD_NOT_FOUND)
+
+        upload_workspace = UploadWorkspace(upload_id)
+        checkpoint_list = upload_workspace.list_checkpoints(user)
+
+
+        response_data = {'upload_id': upload_id,
+                         'checkpoints': json.loads(json.dumps(checkpoint_list))
```
erickpeirson, line 1519: Why dump and then load here?

filemanager/controllers/upload.py
```diff
@@ -1319,6 +1337,420 @@ def get_upload_service_log() -> Response:
...
+    upload_id : int
+        The unique identifier for upload workspace.
+    checkpoint_checksum: str
+        Unique identifier/key for checkpoint.
+    user : str
+        User making create checkpoint request.
+
+    Returns
+    -------
+    Standard Response tuple containing response content, HTTP status, and HTTP headers.
```
erickpeirson:
```suggestion
    tuple
        Standard Response tuple containing response content, HTTP status, and 
        HTTP headers.
```

filemanager/controllers/upload.py
```diff
@@ -1319,6 +1337,420 @@ def get_upload_service_log() -> Response:
...
+
+    """
+    user_id = str(user.username)
```
erickpeirson:
```suggestion
    user_id = str(user.user_id)
```

filemanager/controllers/upload.py
```diff
@@ -1319,6 +1337,420 @@ def get_upload_service_log() -> Response:
...
+        upload_workspace = UploadWorkspace(upload_id)
+        result = upload_workspace.restore_checkpoint(checkpoint_checksum, user)
+
+
+        response_data = {'reason': f"Restored checkpoint '{checkpoint_checksum}'"}  # Get rid of pylint error
+        status_code = status.OK
+
+    except IOError:
+        logger.error("%s: Restore checkpoint request failed ", upload_id)
+        raise InternalServerError(CANT_DELETE_FILE)
```
erickpeirson, line 1584: Right error?

filemanager/controllers/upload.py
```diff
@@ -1319,6 +1337,420 @@ def get_upload_service_log() -> Response:
...
+    upload_id : int
+        The unique identifier for upload workspace.
+    checkpoint_checksum: str
+        Unique identifier/key for checkpoint to delete.
+    user : str
+        User making create checkpoint request.
+
+    Returns
+    -------
+    Standard Response tuple containing response content, HTTP status, and HTTP headers.
```
erickpeirson:
```suggestion
    tuple
        Standard Response tuple containing response content, HTTP status, and 
        HTTP headers.
```

filemanager/controllers/upload.py
```diff
@@ -1319,6 +1337,420 @@ def get_upload_service_log() -> Response:
...
+
+    """
+    user_id = str(user.username)
```
erickpeirson:
```suggestion
    user_id = str(user.user_id)
```

filemanager/controllers/upload.py
```diff
@@ -1319,6 +1337,420 @@ def get_upload_service_log() -> Response:
...
+        upload_workspace = UploadWorkspace(upload_id)
+        result = upload_workspace.delete_checkpoint(checkpoint_checksum, user)
+
+
+        response_data = {'reason': f"Deleted checkpoint '{checkpoint_checksum}'"}  # Get rid of pylint error
+        status_code = status.OK
+
+    except IOError:
+        logger.error("%s: Deleted checkpoint request failed ", upload_id)
+        raise InternalServerError(CANT_DELETE_FILE)
```
erickpeirson, line 1640: Right error?

filemanager/controllers/upload.py
```diff
@@ -1319,6 +1337,420 @@ def get_upload_service_log() -> Response:
...
+    Parameters
+    ----------
+    upload_id : int
+        The unique identifier for upload workspace.
+    user : str
+        User making create checkpoint request.
+
+    Returns
+    -------
+    Standard Response tuple containing response content, HTTP status, and HTTP headers.
```
erickpeirson:
```suggestion
    tuple
        Standard Response tuple containing response content, HTTP status, and 
        HTTP headers.
```

filemanager/controllers/upload.py
```diff
@@ -1319,6 +1337,420 @@ def get_upload_service_log() -> Response:
...
+
+    """
+    user_id = str(user.username)
```
erickpeirson:
```suggestion
    user_id = str(user.user_id)
```

filemanager/controllers/upload.py
```diff
@@ -1319,6 +1337,420 @@ def get_upload_service_log() -> Response:
...
+        upload_workspace = UploadWorkspace(upload_id)
+        checkpoint_list = upload_workspace.delete_all_checkpoints(user)
+
+
+        response_data = {'reason': f"Deleted all checkpoints."}  # Get rid of pylint error
+        status_code = status.OK
+
+    except IOError:
+        logger.error("%s: Delete all checkpoints request failed ", upload_id)
+        raise InternalServerError(CANT_DELETE_FILE)
```
erickpeirson: Right error?

filemanager/controllers/upload.py
```diff
@@ -1319,6 +1337,420 @@ def get_upload_service_log() -> Response:
...
+    Parameters
+    ----------
+    upload_id : int
+        The unique identifier for upload workspace.
+    checkpoint_checksum: str
+        checksum of checkpoint ile to be checked.
+
+    Returns
+    -------
+    Standard Response tuple containing content, HTTP status, and HTTP headers.
```
erickpeirson:
```suggestion
    tuple
        Standard Response tuple containing content, HTTP status, and 
        HTTP headers.
```

filemanager/controllers/upload.py
```diff
@@ -1319,6 +1337,420 @@ def get_upload_service_log() -> Response:
...
+            return {}, status.OK, {'ETag': checkpoint_checksum,
+                                   'Content-Length': size,
+                                   'Last-Modified': modified
+                                   }
+
+        raise NotFound(f"Checksum '{checkpoint_checksum}' not found.")
+
+    except IOError:
+        logger.error("%s: Checkpoint file exists request failed ", upload_db_data.upload_id)
+        raise InternalServerError(CANT_DELETE_FILE)
```
erickpeirson, line 1745: Right error?

filemanager/process/tests/test_process_upload.py
```diff
@@ -212,6 +212,39 @@
...
+    Parameters
+    ----------
+    list1: list
+        List of file objects
+    list2
+        List of file objects
+    Returns
+    -------
+    True if list of File objects is idenical on name, size, checsum fields,
+    otherwise False is returned.
```
erickpeirson, line 233:
```suggestion
    bool
        True if list of File objects is identical on name, size, checksum fields,
        otherwise False is returned.
```

filemanager/process/tests/test_process_upload.py
```diff
@@ -212,6 +212,39 @@
...
+
+    """
+    names1 = {file.name for file in list1}
```
erickpeirson, line 236: Avoid using the built-in name ``file``. Even this would work:
```suggestion
    names1 = {file_.name for file_ in list1}
```

filemanager/process/tests/test_process_upload.py
```diff
@@ -212,6 +212,39 @@
...
+    names2 = {file.name for file in list2}
+
+    if len(names1) != len(names2):
+        return False
+
+    # Compare the two lists
+    set1 = set((x.name, x.checksum, x.size) for x in list1)
+
+    difference = [x for x in list2 if (x.name, x.checksum, x.size) not in set1]
```
erickpeirson, line 245: Why not use set comparisons here as well?

```python
set1 = {(f.name, f.checksum, f.size) for f in list1}
set2 = {(f.name, f.checksum, f.size) for f in list2}
if not set1 ^ set2:   #  The symmetric difference between the two sets.
    return True
return False
```

filemanager/process/tests/test_process_upload.py
```diff
@@ -212,6 +212,39 @@
...
+    Returns
+    -------
+    True if list of File objects is idenical on name, size, checsum fields,
+    otherwise False is returned.
+
+    """
+    names1 = {file.name for file in list1}
+    names2 = {file.name for file in list2}
+
+    if len(names1) != len(names2):
```
erickpeirson, line 239: Seems like this comparison is subsumed by the comparison below?

filemanager/process/tests/test_process_upload.py
```diff
@@ -212,6 +212,39 @@
                     ])
 
 
+def compare_file_lists(list1: list, list2: list) -> bool:
```
erickpeirson, line 218: Seems like this name kind of obscures the semantics, i.e. it's not obvious what ``compare_file_lists(..) == True`` implies. Consider ``file_lists_are_equivalent(...)`` or similar?

filemanager/process/upload.py
```diff
@@ -2834,3 +2850,345 @@ def source_format(self) -> str:
...
+        # Find the checkpoint file
+        for root_directory, directories, files in os.walk(checkpoint_directory):
+            for file in files:
+                path = os.path.join(root_directory, file)
+                obj = File(path, checkpoint_directory)
+
+                if obj.checksum == checkpoint_checksum:
+                    return obj
+
+        return None
```
erickpeirson: Consider raising an exception instead of returning ``None``. Keeps your typing simpler, forces the caller to explicitly contend with the case there is no matching checkpoint.

But if you do keep the null return, be sure to update the return type of the method to ``Optional[str]``.

filemanager/process/upload.py
```diff
@@ -2834,3 +2850,345 @@ def source_format(self) -> str:
...
+
+
+    def checkpoint_file_path(self, checkpoint_checksum: str) -> str:
```
erickpeirson: Consider adding a verb to the method name, e.g. ``get_checkpoint_file_path(...)`` -- since ``checkpoint`` could be read as a verb it's a little ambiguous.

filemanager/process/upload.py
```diff
@@ -2834,3 +2850,345 @@ def source_format(self) -> str:
...
+        Path to checkpoint specified by unique checksum.
+        Null if file does not exist.
+
+        """
+        file_obj = self.resolve_checkpoint_file_obj(checkpoint_checksum)
+
+        if file_obj is not None:
+            return file_obj.filepath
+
+        return ""
```
erickpeirson: Recommend raising an exception here instead of returning an empty string

filemanager/process/upload.py
```diff
@@ -2834,3 +2850,345 @@ def source_format(self) -> str:
...
+
+        Parameters
+        ----------
+        checkpoint_checksum : str
+            Checksum that uniquely identifies checkpoint.
+
+        Returns
+        -------
+        Path to checkpoint specified by unique checksum.
+        Null if file does not exist.
```
erickpeirson, line 2910:
```suggestion
        str
            Path to checkpoint specified by unique checksum.
```

filemanager/process/upload.py
```diff
@@ -2834,3 +2850,345 @@ def source_format(self) -> str:
...
+        Indicate whether checkpoint files exists.
+
+        Parameters
+        ----------
+        checkpoint_checksum : str
+            Checksum that uniquely identifies checkpoint.
+
+        Returns
+        -------
+        True if file exists, False otherwise.
```
erickpeirson:
```suggestion
        bool
            True if file exists, False otherwise.
```

filemanager/process/upload.py
```diff
@@ -2834,3 +2850,345 @@ def source_format(self) -> str:
...
+
+        """
+        file_obj = self.resolve_checkpoint_file_obj(checkpoint_checksum)
+
+        if file_obj is not None:
+            return os.path.exists(file_obj.filepath)
+
+        return False
+
+    def checkpoint_file_size(self, checkpoint_checksum: str) -> int:
```
erickpeirson: Recommend adding a verb, i.e. ``checkpoint_file_size(...)``

filemanager/process/upload.py
```diff
@@ -2834,3 +2850,345 @@ def source_format(self) -> str:
...
+
+        Returns
+        -------
+        Size in bytes.
+        """
+        file_obj = self.resolve_checkpoint_file_obj(checkpoint_checksum)
+        if file_obj is not None:
+            return file_obj.size
+
+        return 0
```
erickpeirson: Recommend raising an exception. This could be construed as implying that the file exists, and is 0 bytes in size.

filemanager/process/upload.py
```diff
@@ -2834,3 +2850,345 @@ def source_format(self) -> str:
...
+        -------
+        File pointer or Null string when filepath does not exist.
+
+        """
+        file_obj = self.resolve_checkpoint_file_obj(checkpoint_checksum)
+
+        if file_obj is not None and os.path.exists(file_obj.filepath):
+            return open(file_obj.filepath, 'rb')
+
+        return ""
```
erickpeirson: Raise exception. If the caller expects an ``io.BytesIO`` this will be really confusing

filemanager/process/upload.py
```diff
@@ -2834,3 +2850,345 @@ def source_format(self) -> str:
...
+        -------
+        Size in bytes.
+        """
+        file_obj = self.resolve_checkpoint_file_obj(checkpoint_checksum)
+        if file_obj is not None:
+            return file_obj.size
+
+        return 0
+
+    def checkpoint_file_pointer(self, checkpoint_checksum: str) -> io.BytesIO:
```
erickpeirson: Recommend adding a verb

erickpeirson (commented):
Awesome to see this implemented! Didn't get all the way through the PR yet, but mostly find myself commenting on the same set of things:
- Should probably use user_id instead of username
- Use exceptions instead of null returns
- Docstring formatting
- Look for opportunities to improve semantics -- method names without verbs can be a little ambiguous in some cases.

Responses:

- Should probably use user_id instead of username. I forget username is not immutable. I will use the user_id.
- Use exceptions instead of null returns. I will work to correct this bad habit.
- Docstring formatting. I will review docstring formatting and make suggested changes.
- Look for opportunities to improve semantics -- method names without verbs can be a little ambiguous in some cases. I will review method names.

filemanager/controllers/upload.py
DavidLFielding: In this case, I am indicating username as 'NA' for logging purposes. None will also work so I will use None instead.

filemanager/controllers/upload.py
DavidLFielding: After discussing with Jake I believe we need an optional argument to clear out source files after checkpoint and before unpacking uploaded files.

I will make this an argument.

filemanager/controllers/upload.py
DavidLFielding: Fixed 15 occurrences.

filemanager/controllers/upload.py
DavidLFielding: # Admin may want to clear out existing files so that existing files
        # do not cause problems or interfere with processing uploaded files.
        #
        # Steps are 1) checkpoint 2) clear source files 3) unpack upload
        #
        if clear_source_files:
            upload_workspace.client_remove_all_files()

erickpeirson: Awesome; much clearer

filemanager/controllers/upload.py
DavidLFielding: This makes a lot of sense and will eliminate duplicate code that raises an exception. I've tended to surround multiple methods called during requests by a single try/except. For example, in restore checkpoint the methods  uploads.retrieve and upload_workspace.restore_checkpoint both have the possibility of generating not found exception (or they will soon). The single try/except seems less clunky but if you prefer I will surround each with try/except.

DavidLFielding: I'm looking at my exception handling a little more and examining exception handling across other arXiv repositories.

In the above case wouldn't raising a NotFound exception in uploads.retrieve be sufficient to flag, uniquely identify, and log the 'upload id not found' exception given that the enclosing except clause in controller already checks for the NotFound exception and reraises it?

This one line change (uploads.retrieve) would allow me to get rid of the repetitive code in the controller that checks for 'None' (upload id not found in DB) and raises a NotFound exception. A definite improvement.

I scanned across all of the arXiv repositories that I have checked out (most of them) and I discovered only a handful of custom exceptions. This leads me to ask whether general/werkzeug exceptions cover most of what we need or do we need to define more custom exceptions. 

I'd like to suggest a broader discussion on the desire/need for custom exceptions. 

Sorry to go further off scope but if I'm off on my exception handling I'd like to correct it sooner than later.

filemanager/process/upload.py
DavidLFielding: Fixed.

filemanager/controllers/upload.py
DavidLFielding: I'm mainly using user information to log who is making a particular request. I understand that user_id in immutable. I would argue that being able to look at the log and glean some information about who is making a particular request is what I'm really after. Logging the user_id alone seems to lose part of what I'm trying to accomplish.

14/May/2019:14:16:59 -0400 Creating checkpoint. ['1']

so I'm wondering if something like the following might be more useful:

[user:\<user_id>\<username>] such that log entry looks like:

14/May/2019:14:16:59 -0400 Creating checkpoint. [user:1:johnny]

This would be useful to someone looking at log and easily parsable in the event we want to do further processing (like display full name).

filemanager/process/upload.py
DavidLFielding: Now: raise NotFound(CHECKPOINT_FILE_NOT_FOUND)

filemanager/process/upload.py
DavidLFielding: Recommend adding a verb, i.e. checkpoint_file_size(...)

My eyes must be deceiving me here. Do you mean like get_checkpoint_file_size()?

filemanager/process/upload.py
DavidLFielding: Fixed.

filemanager/process/upload.py
erickpeirson: Oh dear, that's what I get for keeping odd hours. Yes, that's what I meant

filemanager/controllers/upload.py
erickpeirson: > In the above case wouldn't raising a NotFound exception in uploads.retrieve be sufficient to flag, uniquely identify, and log the 'upload id not found' exception given that the enclosing except clause in controller already checks for the NotFound exception and reraises it?

In this case I would argue that we want to maintain separation of concerns between the controllers and the service module. The service module shouldn't have to know anything about HTTP and 404 Not Found (which ``werkzeug.exceptions.NotFound`` represents). Another way to say this is to observe that the concept of a resource not being found in HTTP is not the same concept as a record not being found in a database.

> I scanned across all of the arXiv repositories that I have checked out (most of them) and I discovered only a handful of custom exceptions. This leads me to ask whether general/werkzeug exceptions cover most of what we need or do we need to define more custom exceptions.

We shouldn't need custom HTTP exceptions. But we should be open to custom exceptions that express semantically relevant information. A record not being found seems appropriate.

> I'd like to suggest a broader discussion on the desire/need for custom exceptions.

Absolutely! Will whip up a brief description of what I have in mind, and maybe we can use that as a discussion piece for the team.

filemanager/controllers/upload.py
erickpeirson: > so I'm wondering if something like the following might be more useful:
> 
> [user:1:johnny] such that log entry looks like:
> 
> 14/May/2019:14:16:59 -0400 Creating checkpoint. [user:1:johnny]

That looks good to me

filemanager/process/upload.py
DavidLFielding: I followed your lead and attempted to name checkpoint methods in a consistent way to the similar content methods you originally created:

    def content_file_path(self, public_file_path: str) -> str:
    def content_file_exists(self, public_file_path: str) -> bool:
    def content_file_size(self, public_file_path: str) -> int:
    def content_file_checksum(self, public_file_path: str) -> str:
    def content_file_pointer(self, public_file_path: str) -> io.BytesIO:
    def content_file_last_modified(self, public_file_path: str) -> datetime:

It seems the "add verb" recommendation applies to a number of methods. I agree that adding a verb is helpful and will go ahead and bring method names up to spec.

erickpeirson: Cool, that looks reasonable to me

filemanager/controllers/upload.py
DavidLFielding: Improved on user logging with [user:<user_id><username>] format. I also added user logging to other requests that modify workspace.

filemanager/process/upload.py
DavidLFielding: I tidied up the method names a bit. We could still use some work on predicates after we discuss predicates decision proposal. Seems methods like "content_file_exists" might be improved by name like "does_content_file_exists".

erickpeirson: Or even ``has_content_file``. But what you have looks good

erickpeirson:
Closing this in favor of #43

PR #43, opened 2019-06-05
ARXIVNG-2210 ARXIVNG-2211 filemanager refactor

erickpeirson:
The initial implementation of the file manager service focused on replicating  the behavior of the legacy file checks, and so closely reproduced much of the  logic and abstractions. This implementation was quite fast in local tests, correct, and had excellent test coverage over known problematic uploads.

### Problem: EFS I/O is too slow to run checks

EFS I/O limits scale with overall volume size, which makes things difficult from the start. But even if we provision a high level of I/O ahead of time,  there are too many low-level reads and writes to perform file checks on EFS  directly. A modestly sized tarball could take around 12 seconds to unpack and check, which is not acceptable.

So we decided that we need to refactor the application to perform checks on a local disk (e.g. in memory), and then move files over to EFS at the end of the request. This way we are keeping I/O to a minimum.

### Problem: the core checking and file-shuffling routines were too complex

Because the initial implementation stuck closely to the logic of the legacy  system, the file management and checking logic had grown to nearly 3,000 lines largely contained in a single class, with really long methods doing the  actual checking. This made it hard to understand what checks were being  performed, and make changes that were predictable.

Since we were going to refactor to address the I/O issues, we decided to go  ahead and perform an initial refactor for clarity, composability, and extensibility.

### Refactor

Here is an overview of the refactor in 2019-06:

- Abstracted away the underlying filesystem logic by encapsulating it in
  adapters. A  [``SimpleStorageAdapter``](https://github.com/arXiv/arxiv-filemanager/blob/42b9e69eb24d5f41c5eb1667cb1215fa7b83484b/filemanager/services/storage.py#L24)  implements logic that is close to the original implementation (on a single  volume). A  [``QuarantineStorageAdapter``](https://github.com/arXiv/arxiv-filemanager/blob/42b9e69eb24d5f41c5eb1667cb1215fa7b83484b/filemanager/services/storage.py#L271)  implements the two-volume logic needed to do fast checks before shipping data  to EFS. The storage adapters implement an API/protocol that is [formalized in  the domain](https://github.com/arXiv/arxiv-filemanager/blob/42b9e69eb24d5f41c5eb1667cb1215fa7b83484b/filemanager/domain/storage.py#L1).

- Separated the file/workspace checks from the upload workspace class itself using multiple-dispatch (something like the visitor pattern). The advantage of this pattern is that we can add more checks, reorder the checks, etc, without creating an even more enormous class. We can also test them   separately, which we should start to do...
  
  - Checkers (visitors) are [formalized in the domain](https://github.com/arXiv/arxiv-filemanager/blob/42b9e69eb24d5f41c5eb1667cb1215fa7b83484b/filemanager/domain/checks.py#L1), and extend a [``BaseChecker`` class](https://github.com/arXiv/arxiv-filemanager/blob/42b9e69eb24d5f41c5eb1667cb1215fa7b83484b/filemanager/process/check/base.py#L21).
  - Checking strategies are also [formalized in the  domain](https://github.com/arXiv/arxiv-filemanager/blob/42b9e69eb24d5f41c5eb1667cb1215fa7b83484b/filemanager/domain/checks.py#L15),  and implement the visitation of the checkers on a workspace. The [``SynchronousCheckingStrategy``](https://github.com/arXiv/arxiv-filemanager/blob/42b9e69eb24d5f41c5eb1667cb1215fa7b83484b/filemanager/process/strategy.py#L13)
    is the first such implementation.
  - The ``UploadWorkspace`` class is assigned a [set of checks and a  strategy](https://github.com/arXiv/arxiv-filemanager/blob/42b9e69eb24d5f41c5eb1667cb1215fa7b83484b/filemanager/domain/uploads.py#L758-L762), which are applied via the [``UploadWorkspace.perform_checks``](https://github.com/arXiv/arxiv-filemanager/blob/42b9e69eb24d5f41c5eb1667cb1215fa7b83484b/filemanager/domain/uploads.py#L692) method.

- For clarity sake, decomposed most of the ``UploadWorkspace`` into [mixins that add sets of
  functionality](https://github.com/arXiv/arxiv-filemanager/blob/42b9e69eb24d5f41c5eb1667cb1215fa7b83484b/filemanager/domain/uploads.py#L1).
- Consolidated the storage of metadata about the workspace. 

  - Previously, details about the workspace state, and about the files in the workspace, were generated on each request (indeed, all or most checks were re-run on each request). In addition, a database row was maintained with some metdata about the last run through the checks. 
  - In the reimplemented checks, file-level checks are only run once per file, and workspace-level checks are only run after files are added or deleted. 
  - In the new implementation, the database service has the primary responsibility for loading metadata about the workspace, and initializing the ``UploadWorkspace`` with its storage adapter. This means that loading the workspace requires only a single function call, and there is no need to keep track of both a filesystem-based workspace object and a database object (both of which represent the workspace) in the controllers.

- Since the core classes were significantly altered, much of the work involved modifying the extensive test suite to use the new internal APIs, and  modifying the controllers to leverage the new patterns in the domain and processes. Since this involved going through the test suite one case at a time, I took the liberty of breaking things up into more manageable pieces:

  - Split controllers out into submodules, for easier navigation.
  - Split up super long test routines in the API tests into separate modules in
    (``tests/test_api/``), smaller ``TestCase``s, and smaller ``test_`` 
    methods. Hopefully this makes it easier to find relevant tests and also see
    what specifically is being tested.

- Removed some cruft that wasn't being used and wasn't likely to be used
  (async boilerplate, etc).


DavidLFielding (approved):
The small number of changes that allow you to authenticate with user or client credentials looks reasonable. 

Tests fail for me but your note seems to indicate you are already aware of this.

I haven't looked at your rewrite of the file check routines in any depth as these do not appear to be part of this PR. I'm not really sure where to start on those changes.

erickpeirson:
Added documentation: https://arxiv.github.io/arxiv-filemanager/

erickpeirson:
This branch on submission UI has up to date refs to work with this branch of filemanager: https://github.com/arXiv/arxiv-submission-ui/tree/task/filemanager-refactor

erickpeirson:
@mhl10 @DavidLFielding had to make some changes to get static checks to work correctly. docstyle and linter may still complain

.pytest_cache/v/cache/lastfailed
```diff
@@ -0,0 +1,37 @@
+{
```
bdc34: It seems like .pytest_cache shouldn't be checked into git.

erickpeirson: yikes indeed

erickpeirson: fixed in ea8c33f

Pipfile
```diff
@@ -5,23 +5,25 @@ name = "pypi"
 
 [packages]
 arxiv-base = "==0.15.8rc4"
```
mhl10:
```suggestion
arxiv-base = "==0.15.9"
```

Pipfile
```diff
@@ -5,23 +5,25 @@ name = "pypi"
...
-jsonschema = "*"
-celery = "*"
-redis = "*"
-arxiv-auth = "==0.3.2rc6"
+flask = "==1.0.3"
+flask-sqlalchemy = "==2.4.0"
+requests = "==2.22.0"
+uwsgi = "==2.0.18"
+jsonschema = "==3.0.1"
+arxiv-auth = "==0.3.4rc1"
```
mhl10:
```suggestion
arxiv-auth = "==0.4.1"
```

filemanager/controllers/_messages.py
```diff
@@ -0,0 +1,58 @@
...
+
+# exceptions
+UPLOAD_MISSING_FILE = 'missing file/archive payload'
+UPLOAD_MISSING_FILENAME = 'file argument missing filename or file not selected'
+UPLOAD_FILE_EMPTY = 'file payload is zero length'
+
+UPLOAD_NOT_FOUND = 'workspace not found'
+UPLOAD_CHECKPOINT_NOT_FOUND = 'checkpoint not found'
+UPLOAD_DB_ERROR = 'unable to create/insert new upload workspace into database'
+UPLOAD_DB_CONNECT_ERROR = "There was a problem connecting to database."
```
mhl10:
```suggestion
UPLOAD_DB_CONNECT_ERROR = 'There was a problem connecting to the database.'
```

filemanager/controllers/_messages.py
```diff
@@ -0,0 +1,58 @@
...
+UPLOAD_DELETED_ALL_CHECKPOINTS = 'deleted all checkpoint'
+UPLOAD_WORKSPACE_IS_EMPTY = 'workspace is empty'
+
+UPLOAD_NOT_ACTIVE = 'upload workspace is not active.'
+UPLOAD_WORKSPACE_LOCKED = 'upload workspace is locked.'
+
+UPLOAD_WORKSPACE_ALREADY_DELETED = 'Request failed. Workspace has been deleted.'
+
+# upload status codes
+# INVALID_UPLOAD_ID = {'reason': 'invalid upload identifier'}
```
mhl10, line 35: Not used anymore?

filemanager/controllers/checkpoint.py
```diff
@@ -0,0 +1,413 @@
...
+    """
+    user_string = util.format_user_information_for_logging(user)
+    logger.info("%s: Create checkpoint [%s].", upload_id, user_string)
+
+    try:
+        # Make sure we have an upload_db_data to work with
+        workspace = database.retrieve(upload_id)
+        checksum = workspace.create_checkpoint(user)
+
+        ###
```
mhl10, line 47: Unused code

filemanager/controllers/checkpoint.py
```diff
@@ -0,0 +1,413 @@
...
+        raise httpe
+    except FileNotFoundError as fne:
+        raise NotFound(f"Checksum '{checkpoint_checksum}' not found.") from fne
+    except IOError as ioe:
+        logger.error("%s: Checkpoint file exists request failed ",
+                     upload_id)
+        raise InternalServerError(messages.CANT_DELETE_FILE) from ioe
+    except UploadFileSecurityError as secerr:
+        logger.info("%s: %s", upload_id, secerr)
+        # TODO: Should this be BadRequest or NotFound. I'm leaning towards
```
mhl10, line 340: `BadRequest` seems fine here.

filemanager/controllers/files.py
```diff
@@ -0,0 +1,310 @@
...
+                     workspace.upload_id)
+        raise InternalServerError(messages.CANT_DELETE_FILE)
+    except SecurityError as secerr:
+        logger.info("%s: %s", upload_id, secerr.description)
+        # TODO: Should this be BadRequest or NotFound. I'm leaning towards
+        # NotFound in order to provide as little feedback as posible to client.
+        raise NotFound(messages.UPLOAD_FILE_NOT_FOUND)
+    except Exception as ue:
+        logger.error("Unknown error in content file exists operation. "
+                    " Add except clauses for '%s'. DO IT NOW!", ue)
```
mhl10, line 83: Hoping this doesn't get lost in the flood of logs.

filemanager/controllers/files.py
```diff
@@ -0,0 +1,310 @@
...
+        # propagate.
+        logger.info("%s: Operation failed: '%s'.", httpe, upload_id)
+        raise httpe
+    except database.WorkspaceNotFound as nf:
+        logger.info("%s: Workspace not found: '%s'", upload_id, nf)
+        raise NotFound(messages.UPLOAD_NOT_FOUND)
+
+    except IOError:
+        raise
+        # logger.error("%s: Delete all files request failed ", upload_id)
```
mhl10, line 292: It's probably still worth logging this before the `raise`.

filemanager/controllers/files.py
```diff
@@ -0,0 +1,310 @@
...
+    int
+        HTTP status
+    dict
+        HTTP headers.
+
+    """
+    try:
+        workspace: UploadWorkspace = database.retrieve(upload_id)
+    except IOError:
+        logger.error("%s: ContentFileExistsCheck: There was a problem "
```
mhl10, line 54: Log message can be reused from `messages.UPLOAD_DB_CONNECT_ERROR` to avoid duplication, and to be more consistent with other parts of this code.

filemanager/controllers/release.py
```diff
@@ -0,0 +1,180 @@
...
+        if workspace is None:
+            # Invalid workspace identifier
+            raise NotFound(messages.UPLOAD_NOT_FOUND)
+
+        if workspace.is_released:
+            logger.info("%s: Release: Workspace has already been released "
+                        "[%s].", upload_id, user_string)
+            response_data = {'reason': messages.UPLOAD_RELEASED_WORKSPACE}
+
+            status_code = status.OK # Should this be an error?
```
mhl10, line 69: I can't think of any reason it should be an error.

filemanager/controllers/service_log.py
```diff
@@ -0,0 +1,158 @@
...
+    ----
+    Service log should exist. Response includes size of log and last
+    modified date which may be useful for inquiries where client does
+    not desire to download entire log file.
+
+    Returns
+    -------
+    Standard Response tuple containing content, HTTP status, and HTTP headers.
+    """
+    # Need path to upload.log which is currently stored at top level
```
mhl10, line 113: Comment seems redundant.

filemanager/domain/file_type.py
```diff
@@ -0,0 +1,166 @@
...
+        Indicate whether this is a TeX type.
+
+        This method does some normalization prior to calling internal routine.
+        """
+        return self in TEX_TYPES
+
+    # QUESTION: The only place that I see this logic used is in tests. Is it
+    # actually still needed? -- Erick 2019-06-10
+    @property
+    def priority(self) -> int:
```
mhl10, line 85: It doesn't look like this is being used anywhere, but @DavidLFielding can confirm whether it is/will be important.

filemanager/domain/uploaded_file.py
```diff
@@ -0,0 +1,172 @@
...
+        return os.path.splitext(self.path)[0]
+
+    @property
+    def ext(self) -> str:
+        """Return file extension."""
+        _, ext = os.path.splitext(self.path)
+        return ext.lstrip('.')
+
+    @property
+    def dir(self) -> str:
```
mhl10, line 130: Consider renaming this property because it shares its name with a built-in.

filemanager/domain/uploads/checkpoint.py
```diff
@@ -0,0 +1,312 @@
...
+
+        """
+        u_file = self._resolve_checkpoint_file(checksum)
+        if u_file is not None:
+            return u_file.path
+
+        raise FileNotFoundError(UPLOAD_FILE_NOT_FOUND)
+
+
+    def checkpoint_file_exists(self, checksum: str) -> bool:
```
mhl10, line 91:
```suggestion
    def does_checkpoint_file_exist(self, checksum: str) -> bool:
```

mhl10 (approved):
Looks good overall! Nice documentation, ample tests. Consider a quick review of `TODO`s since there's a fairly large number.

erickpeirson:
> Looks good overall! Nice documentation, ample tests. Consider a quick review of `TODO`s since there's a fairly large number.

Awesome, thanks! Some of these are from David's original code, so I was ambivalent about how much pruning to do. Will take a look

DavidLFielding:
I'm really just getting started on my review and am currently taking a leisure look over the documentation and code. The PR writeup and architecture documentation is very helpful. I ran the tests and they completed successfully. Lots of tests! Looks like you may have even added more!

I attempted to experiment with the development server per the README and I suspect this documentation needs to be updated. The populate_test_database.py fails to run after I performed the pipenv install and pipenv shell commands. Does this need to be updated for the refactored service?

The command to start up the development server worked fine for me but it doesn't appear to like my Requestly token in Chrome when I execute the rudamentary /status request (http://localhost:5000/filemanager/api/status). I generated a new token that appears to have upload:admin scope yet I'm still getting an error.  I don't remember status request being limited by scope. Looking at routes upload_api.py file the /status request doesn't appear to have a scope limiting client access. That's different. I don't want to get bogged down here so I'll follow up on this later after I spend more time reviewing the code.

erickpeirson:
Updated start-up instructions and added info about auth tokens to README.md in 8c046d6

filemanager/controllers/files.py
```diff
@@ -0,0 +1,310 @@
...
+                            user: auth_domain.User) -> Response:
+    """
+    Get the source log associated with upload workspace.
+
+    Parameters
+    ----------
+    upload_id : int
+        The unique identifier for upload workspace.
+    public_file_path: str
+        relative path of file to be deleted.
```
DavidLFielding, line 103: The relative path of the file to be retrieved.

filemanager/controllers/files.py
```diff
@@ -0,0 +1,310 @@
...
+               'ETag': u_file.checksum,
+               'Content-Length': u_file.size_bytes,
+               'Last-Modified': u_file.last_modified}
+    return {}, status.OK, headers
+
+
+def get_upload_file_content(upload_id: int, public_file_path: str,
+                            user: auth_domain.User) -> Response:
+    """
+    Get the source log associated with upload workspace.
```
DavidLFielding, line 96: Retrieve a single content file from the workspace.

filemanager/controllers/files.py
```diff
@@ -0,0 +1,310 @@
...
+    Delete a single file.
+
+    This request is being received from API so we need to be extra careful.
+
+    Parameters
+    ----------
+    upload_id : int
+        The unique identifier for the workspace in question.
+    public_file_path: str
+        relative path of file to be deleted.
```
DavidLFielding, line 173: The relative path of file to be deleted.
or
Relative path of file to be deleted.

filemanager/controllers/files.py
```diff
@@ -0,0 +1,310 @@
...
+    """
+    Delete all files uploaded by client from specified workspace.
+
+    This request is being received from API so we need to be extra careful.
+
+    Parameters
+    ----------
+    upload_id : int
+        The unique identifier for the workspace in question.
+    public_file_path: str
```
DavidLFielding, line 253: Don't need public_file_path as argument to delete_all_files.

Do we need to add description for _user: auth_domain.User_ to parameter for all requests? I believe this parameter was a recent addition in preparation for admin operations on workspace.

filemanager/controllers/release.py
```diff
@@ -0,0 +1,180 @@
...
+Response = Tuple[Optional[dict], status, dict]
+
+
+@database.atomic
+def upload_release(upload_id: int, user: auth_domain.User) -> Response:
+    """
+    Release inidcates owner is done with upload workspace.
+
+    System will schedule to remove files.
+
```
DavidLFielding, line 34: This might be clearer with something like:

The system will schedule workspace for (future) deletion. This includes user-uploaded content files and system-generated files [checkpoints, backups, removed files/archives, etc.]).

filemanager/controllers/release.py
DavidLFielding: I most likely wrote this. I suspect I was thinking that extra requests to release an already released workspace might indicate a potential problem with inconsistencies in the state of workspace. We can probably delete that comment at this point.

filemanager/controllers/checkpoint.py
```diff
@@ -0,0 +1,413 @@
...
+        -> Response:
+    """
+    Verify that the specified checkpoint content file exists/is available.
+
+    Parameters
+    ----------
+    upload_id : int
+        The unique identifier for upload workspace.
+    checkpoint_checksum: str
+        checksum of checkpoint ile to be checked.
```
DavidLFielding, line 297: The checksum of checkpoint file to be checked.

filemanager/controllers/checkpoint.py
```diff
@@ -0,0 +1,413 @@
...
+    """
+    user_string = util.format_user_information_for_logging(user)
+    logger.info("%s: Delete checkpoint '%s' [%s].", upload_id,
+                checkpoint_checksum, user_string)
+
+    try:
+        # Make sure we have an upload_db_data to work with
+        workspace = database.retrieve(upload_id)
+        workspace.delete_checkpoint(checkpoint_checksum, user)
+        if workspace.source_package.is_stale:
```
DavidLFielding, line 209: I'm not clear how the source package becomes stale following the deletion of a checkpoint file (non-active). Checkpoint archive is independent of current files stored in source directory.

filemanager/controllers/status.py
```diff
@@ -1,20 +1,20 @@
...
+Response = Tuple[Optional[dict], status, dict]
 
 
-def service_status(*args, **kwargs) -> Response:
+def service_status(*args: Any, **kwargs: Any) -> Response:
     """Exercise dependencies and verify operational status."""
     response_data = {}
     response_data['database'] = database.is_available()
-    response_data['filesystem'] = filesystem.is_available()
+    # response_data['filesystem'] = filesystem.is_available()
```
DavidLFielding, line 17: I understand the database now holds all of the workspace metadata but isn't the existence of the workspace in the actual filesystem necessary to complete a reliable 'service is available' attestation? I suppose we are able to complete numerous metadata type requests even after a catastrophic filesystem failure.

filemanager/domain/error.py
```diff
@@ -0,0 +1,41 @@
...
+    severity: Severity
+    """Severity level of this error."""
+
+    message: str
+    """Human-readable error message."""
+
+    code: Code = field(default=Code.UNKNOWN)
+    """Specific code for this error. Not implemented."""
+
+    path: Optional[str] = field(default=None)
```
DavidLFielding, line 37: I'm still reviewing file by file and haven't investigated path usage beyond this file. Is 'path' the absolute path of the file experiencing an error? Or the relative path? Description and context make it sound like an absolute path. I'm only asking because we never display the absolute path of a file to end-user yet we do return errors containing paths (relative) to the client which may end up getting displayed to the user. If there is no restriction on the path is there a mechanism for generating error message with a relative path for end-user context?

erickpeirson: Paths in this context are always relative to the workspace. Mapping to an absolute path is handled in the storage adapter

filemanager/domain/file_type.py
DavidLFielding: AutoTeX uses priority and I expect this to be used in the future when we rewrite AutoTeX in Python. For that reason, I converted all legacy file_type functionality to Python in order to avoid needing to figure out what was omitted at some point in the distant future. I don't believe the few lines of code supporting priority is too unwieldy to leave in for now.

tests/test_files_upload/AfterTermUnPCify.eps
```diff
@@ -30,15 +30,15 @@ rlineto 0 ty neg rlineto closepath fill tox toy moveto 0 setgray show}bind def
 /nl{currentpoint exch pop lmargin exch moveto 0 -10 rmoveto}def/=={/cp 0 def
 typeprint nl}def/typeprint{dup type exec}readonly def/lmargin 72 def/rmargin 72
 def/tprint{dup length cp add rmargin gt{nl/cp 0 def}if dup length cp add/cp
-exch def prnt}readonly def/cvsprint{=string cvs tprint( )tprint}readonly def
+exch def prnt}readonly def/cvsprint{=string cvs tlogger.debug( )tprint}readonly def
```
DavidLFielding, line 33: I'm actually quite surprised to see that the encapsulated postscript reference test files were modified.

erickpeirson: me too; that looks like someone fat-fingered a find/replace... wonder who... ;-)

erickpeirson: should be fixed in fb579d1

deploy/filemanager/templates/10-deployment.yaml
```diff
@@ -81,22 +85,24 @@ spec:
...
             cpu: 100m
-            memory: 128Mi
-        livenessProbe:
-          initialDelaySeconds: 2
-          periodSeconds: 30
-          httpGet:
-            path: "{{ .Values.ingress.path }}/filemanager/api/status"
-            port: 8000
+            memory: 256Mi
+        # livenessProbe:
```
DavidLFielding, line 100: I'm curious why this liveness check has been commented out.

deploy/filemanager/templates/30-ingress.yaml
```diff
@@ -14,6 +14,7 @@ metadata:
     ingress.kubernetes.io/limit-rps: "16"
```
DavidLFielding, line 14: High-level observation: For many of these configuration settings, it's not clear what considerations went into setting these parameters such that other developers might be able to adjust them in the future (aside from the individual setting them). Are certain parameters best guesses or potentially fine-tuned based on live experience from running the system? Are there others that we've had to tweak a lot and need to pay attention to?  I'm assuming 'defaults' are not specified in these files so I'm interested in documentation or comments on how the parameter value was determined.

docs/source/api/filemanager.controllers.rst
```diff
@@ -0,0 +1,29 @@
...
+.. toctree::
+
+    filemanager.controllers.tests
+
+Submodules
+----------
+
+.. toctree::
+
+   filemanager.controllers.files
```
DavidLFielding, line 21: As far as I know these files are generated automatically. Note that the filemanager.controllers.checkpoint (checkpoint.py) is not included here so the docs will need to be regenerated at some point.

filemanager/domain/index.py
```diff
@@ -0,0 +1,97 @@
...
+    Indexing struct for :class:`.UploadedFile`s.
+
+    The overarching objective is to keep track of system, ancillary, removed,
+    and source files without committing to an underlying path/filesystem
+    structure. This helps us maintain flexibility around how we store files.
+    """
+    source: Dict[str, UploadedFile] = field(default_factory=dict)
+    ancillary: Dict[str, UploadedFile] = field(default_factory=dict)
+    removed: Dict[str, UploadedFile] = field(default_factory=dict)
+    system: Dict[str, UploadedFile] = field(default_factory=dict)
```
DavidLFielding, line 26: I take it the system class is for files like checkpoint archives. Would the removed files also fall into the 'system' class of files?

filemanager/domain/tests/test_workspace.py
```diff
@@ -0,0 +1,534 @@
...
+        self.assertEqual(self.mock_storage.move.call_count, 1,
+                         'The underlying storage adapter is called')
+        self.assertTrue(self.wks.exists(self.path), 'Target path still exists')
+        self.assertFalse(self.wks.exists(self.path2),
+                         'Source path does not exist')
+        self.assertEqual(self.wks.get(self.path), self.mock_file2,
+                         'Second file is now at the path of the first file')
+
+    # TODO: implement these.
+    # def test_replace_directory(self):
```
DavidLFielding, line 475: When do these routines need to be implemented by? Does not having this functionality imply we should expect issues or failures from related user behaviors?

filemanager/domain/uploaded_file.py
```diff
@@ -0,0 +1,172 @@
...
+    @property
+    def name(self) -> str:
+        """File name without path/directory info."""
+        if '/' in self.path.strip('/'):
+            basename: str = os.path.basename(self.path)
+            return basename
+        return self.path
+
+    @property
+    def name_sans_ext(self) -> str:
```
DavidLFielding, line 119: Is the same as (os.path.) basename? To be consistent should delete_file() become directory_sans_file()? The method name definitely caught my attention. I'm sure this is just fine.

filemanager/domain/uploads/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+    source upload type.
+  - :class:`.CountableWorkspace`, which adds methods for counting files and
+    types of files.
+
+- :class:`.UploadWorkspace` extends :class:`.CheckableWorkspace` to add an
+  initialization method, and also incorporates:
+
+  - :class:`.ReadinessWorkspace`, which adds semantics around the "readiness"
+    of the workspace for use in a submission.
+  - :class:`.SingleFileWorkspace`, which adds the concept of a "single file
```
DavidLFielding, line 27: We've discussed eventually allowing ancillary files with 'single-file' submissions. Does this implementation limit in any way the ability to provide a single-file submission with ancillary data files?

filemanager/domain/uploads/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+        This is performed on demand, rather than as a ``__post_init__`` hook,
+        so that we have an opportunity to attach an updated :class:`.FileIndex`
+        after the :class:`.UploadWorkspace` is instantiated but before any
+        system files are created.
+        """
+        if self.storage is None:
+            raise RuntimeError('Storage adapter is not set')
+        self.storage.makedirs(self, self.source_path)
+        self.storage.makedirs(self, self.ancillary_path)
+        self.storage.makedirs(self, self.removed_path)
```
DavidLFielding, line 87: Do we need to create system directory to be consistent?

filemanager/domain/uploads/file_mutations.py
```diff
@@ -0,0 +1,421 @@
...
+        for u_file in u_files:
+            parts = u_file.path.split('/')
+            for parent in accumulate(parts[:-1], lambda *p: os.path.join(*p)):
+                parent += '/'
+                if not self.exists(parent):
+                    self.create(parent, FileType.DIRECTORY, is_directory=True)
+            self.files.set(u_file.path, u_file)
+
+    @modifies_workspace()
+    def delete_workspace(self) -> bool:
```
DavidLFielding, line 265: This class contains methods that 'alter files' according to class docstring. The scope of this method seems a bit beyond just altering files. Maybe just inaccurate description or maybe it's in the wrong place.

filemanager/process/check/file_type.py
```diff
@@ -0,0 +1,567 @@
...
+
+def _check_dvips_temp(workspace: CheckableWorkspace,
+                      u_file: UploadedFile) -> Optional[FileType]:
+    """Ignore tmp files created by (unpatched) dvihps, in top dir."""
+    if DVIPS_TEMP_FILE.search(u_file.path):
+        return FileType.ALWAYS_IGNORE    # , '', ''
+    return None
+
+
+# QUESTION: is this still relevant, given that FM service is decoupled from
```
DavidLFielding, line 136: At the current time if you provide the compilation service with a submission containing the missing font file AutoTeX basically aborts. So you could eliminate this check and let the compilation service generate an error. One could argue that many of the checks in the FM service are related to compilation and belong in the compilation service.

filemanager/process/check/missing_references.py
```diff
@@ -0,0 +1,81 @@
...
+import re
+from arxiv.base import logging
+
+from ...domain import FileType, UploadedFile, CheckableWorkspace
+from .base import BaseChecker
+
+logger = logging.getLogger(__name__)
+
+
+class CheckForMissingReferences(BaseChecker):
```
DavidLFielding, line 13: This no longer represents current logic in the Legacy system and needs to be updated.

filemanager/process/check/tex_format.py
```diff
@@ -0,0 +1,34 @@
...
+import re
+from arxiv.base import logging
+
+from ...domain import FileType, UploadedFile, CheckableWorkspace
+from .base import BaseChecker
+
+logger = logging.getLogger(__name__)
+
+
+# TODO: implement this.
```
DavidLFielding, line 13: Looks like you forgot to implement this functionality.

filemanager/process/check/tex_generated.py
```diff
@@ -0,0 +1,56 @@
+"""Check for TeX-generated files."""
```
DavidLFielding, line 1: I'm wondering if this needs to be reworded along the lines of 'eliminate files generated from TeX compilation' to distinguish it more from TeX Produced. TeXGenerated makes me think of TeXProduced.

filemanager/process/check/top_level_directory.py
```diff
@@ -0,0 +1,47 @@
...
+                and entries[0].is_directory and not entries[0].is_ancillary:
+
+            workspace.add_warning(entries[0], "Removed top level directory",
+                                  is_persistant=False)
+            for _, child in workspace.iter_children(entries[0], max_depth=1):
+                _, new_path = child.path.split(entries[0].path, 1)
+                workspace.rename(child, new_path)
+            workspace.remove(entries[0], "Removed top level directory")
+            #
+            # # Set permissions
```
DavidLFielding, line 43: Permissions have always been an issue. I expect you are dealing with these elsewhere.

filemanager/process/check/unpack.py
```diff
@@ -0,0 +1,193 @@
...
+        # ancillary or not.
+        is_ancillary = True if u_file.is_ancillary else None
+
+        full_path = workspace.get_full_path(u_file.dir)
+        zip.extract(zipinfo, full_path)
+        os.utime(full_path)  # Update access and modified times to now.
+        workspace.create(dest, touch=False, is_ancillary=is_ancillary)
+
+
+# TODO: Add support for compressed files.
```
DavidLFielding, line 185: Will we accept Unix compressed files?

filemanager/process/strategy.py
```diff
@@ -0,0 +1,134 @@
...
+from ..domain import CheckableWorkspace, IChecker, ICheckingStrategy, \
+    UploadedFile
+
+logger = logging.getLogger(__name__)
+logger.propagate = False
+
+
+class BaseCheckingStrategy:
+    pass
+    # def _iter_files(self, workspace: 'CheckableWorkspace'):
```
DavidLFielding, line 22: Is this important? I guess not. Should there be more comments here to explain why this code is commented out?

tests/test_api/_tests.py
```diff
@@ -0,0 +1,139 @@
...
+#         """Download an ancillary file."""
+#         response = self.client.get(
+#             f"/filemanager/api/{self.upload_id}/main_a.tex/content",
+#             headers={'Authorization': self.token}
+#         )
+#         self.assertEqual(response.status_code, status.OK)
+#         self.assertIn('ETag', response.headers, "Returns an ETag header")
+
+
+# class TestOneOffSitations(TestCase):
```
DavidLFielding, line 49: I believe this one was more of a template for creating new tests and can probably go.

tests/test_api/_tests.py
```diff
@@ -0,0 +1,139 @@
...
+#                                       headers={'Authorization': admin_token}
+#                                       )
+#
+#         # This cleans out the workspace. Comment out if you want to inspect files
+#         # in workspace. Source log is saved to 'deleted_workspace_logs' directory.
+#         self.assertEqual(response.status_code, status.OK, "Accepted request to delete workspace.")
+#
+#
+#     # Upload a submission package and perform normal operations on upload
+#     def test_upload_files_normal(self) -> None:
```
DavidLFielding, line 109: These normal/typical upload tests were definitely working in the pre-refactored FM service. This test basically scripted a series of typical requests that might be expected during a normal submission process. The idea was to combine different types of requests together in order to catch any inconsistencies in the final workspace state.

tests/test_api/test_ancillary_upload.py
```diff
@@ -0,0 +1,102 @@
...
+from filemanager.services import database
+from filemanager.domain import UploadWorkspace
+
+from .util import generate_token
+
+logger = logging.getLogger(__name__)
+logger.setLevel(int(os.environ.get('LOGLEVEL', '20')))
+
+
+class TestUploadAncillaryFile(TestCase):
```
DavidLFielding, line 32: Handling of ancillary files in incomplete. There was a discussion about implementing a major change in the way we handle ancillary files such that this check code was not completed. If I understand correctly, the checks are completely configurable, so a very limited set of checks needs to be run against the ancillary files. In the legacy system, there were some questions as to whether all of the appropriate checks were being run. The limited checks mainly include the file naming convention checks. All other checks should be ignored for ancillary files (verify this).

tests/test_api/test_checkpoints.py
```diff
@@ -0,0 +1,500 @@
...
+                for chunk in iter(lambda: f.read(4096), b""):
+                    hash_md5.update(chunk)
+            return urlsafe_b64encode(hash_md5.digest()).decode('utf-8')
+
+        return ""
+
+    def unpack_tarfile(self, tarfile_path: str, target_directory: str) -> None:
+        """
+        Fast and dirty routine to unpack
+        Parameters
```
DavidLFielding, line 70: This might look better with a blank line to separate description from parameters.

tests/test_api/test_eps_repair.py
```diff
@@ -0,0 +1,226 @@
...
+        # Compare downloaded file (content_file_path) against reference_path
+        # file.
+        reference_filename = 'dos_eps_1_stripped.eps'
+        reference_path = os.path.join(self.TEST_FILES_STRIP_PS,
+                                      reference_filename)
+        # Compared fixed file to a reference_path stripped version of file.
+        is_same = filecmp.cmp(content_file_path, reference_path, shallow=False)
+        self.assertTrue(is_same, f"Repair EPS file '{test_filename}'.")
+
+#         # Try encapsulate Postscript with trailing TIFF
```
DavidLFielding, line 140: I'm a little worried if we are simply commenting out tests that stop working following the refactor. These are the low-level cleanup routines that strip TIFF images from encapsulated Postscript files. May need to go back and run the pre-refactor versions to verify that they actually worked.

tests/test_api/test_existing_upload.py
```diff
@@ -0,0 +1,101 @@
...
+class TestUploadToExistingWorkspace(TestCase):
+    """
+    Test various failure conditions on existing workspaces.
+
+    The same function handles new upload request so we will not repeat
+    basic argument failures that have been covered already.
+
+    TODO: Add tests for locked/released workspaces on upload requests.
+    TODO: Lock/unlock, release/unrelease. (when implements)
+    TODO: Add size checks (when implemented)
```
DavidLFielding, line 35: I believe size checks are now handled in submission core so you can delete the TODO about size.

tests/test_api/test_feature_missing_references.py
```diff
@@ -0,0 +1,122 @@
...
+from filemanager.services import database
+from filemanager.domain import UploadWorkspace
+
+from .util import generate_token
+
+logger = logging.getLogger(__name__)
+logger.setLevel(int(os.environ.get('LOGLEVEL', '20')))
+
+
+class TestMissingReferences(TestCase):
```
DavidLFielding, line 28: These tests need revision as the bib/bbl logic has changed. (Ignore my comments on bib/bbl handling if you've already updated the internal logic per recent changes)

tests/test_api/test_new_upload.py
```diff
@@ -0,0 +1,262 @@
...
+                         'Status is consistent between requests')
+        self.assertEqual(post_data['checksum'], get_data['checksum'],
+                         'Checksum is consistent between requests')
+        self.assertEqual(len(post_data['files']), len(get_data['files']),
+                         'Number of files is consistent between requests')
+
+        # TODO: need to look a little closer at how we think about readiness.
+        # Right now, a workspace can be READY_WITH_WARNINGS on upload, but then
+        # READY on subsequent status requests. Guess I'm not clear on what
+        # _WITH_WARNINGS should signify -- should it be persistant warnings?
```
DavidLFielding, line 258: A warning should be persistent unless the warning is cleared (user action) or disappears due to system action (ie., a file was deleted). User may choose to ignore warnings but the warnings should not simply disappear on their own.

tests/test_api/test_quarantine.py
```diff
@@ -0,0 +1,344 @@
+"""Tests the application via its API, using the QuarantineStorageAdapter."""
```
DavidLFielding, line 1: This test appears to specifically test the new quarantine storage adaptor. Would it make sense to run all existing tests against each of the storage adaptor instances? Could you set the storage adaptor, run all of the existing tests, set a different storage adaptor, and run all of the tests again? My sense is if we are changing the underlying storage adaptors we should run an exhaustive series of tests that are not specifically tailored to the storage adaptors. Then again we probably need a few tests tailored to the storage adaptors. Just my two cents here.

tests/test_check_workspace.py
```diff
@@ -0,0 +1,953 @@
...
+        self.assertIn(expected_warning,
+                      self.workspace.get_warnings_for_path('PostscriptPreview2.ps'))
+
+        result_path = self.workspace.get_full_path('PostscriptPreview2.ps')
+        expected_path = os.path.join(self.DATA_PATH,
+                                     'PostscriptPreview2_stripped.eps')
+        self.assertTrue(filecmp.cmp(result_path, expected_path, shallow=False),
+                        'Resulting content has preview stripped')
+
+    # QUESTION: here's another weird one; the expected output file has 3322
```
DavidLFielding, line 567: The checks that clean and modify files were traditionally run in a fixed order. So the test reference files are constructed with some knowledge of the order that checks are applied. If you change the order in which checks are applied then it's possible another cleanup routine modified the test file such that it no longer matches the original reference file. I'm just guessing this might be somehow related to this comment about different test output.

tests/test_api/test_upload_session.py
```diff
@@ -0,0 +1,289 @@
...
+from filemanager.services import database
+from filemanager.domain import UploadWorkspace
+
+from .util import generate_token
+
+logger = logging.getLogger(__name__)
+logger.setLevel(int(os.environ.get('LOGLEVEL', '20')))
+
+
+class TestUploadNormalFiles(TestCase):
```
DavidLFielding, line 32: Here's where the normal upload workflow tests ended up. Ignore that previous comment about the commented out version of the normal tests.

DavidLFielding (commented):
I managed to at least glance at most of the files in the PR. Hopefully, my comments are useful in some way. I may check a few more things if I have time but plan to move on to other tasks tomorrow.

At this point, I have not run the service beyond exercising the tests, which all passed.

DavidLFielding:
I notice some formatting issues in my checkpoint PR so on a whim I decided to run some checks on File Manager.

pylint filemanager
...
------------------------------------------------------------------
Your code has been rated at 8.81/10 (previous run: 8.81/10, +0.00)

Yikes! I hear screams off in the distance.

There seem to be a huge number of trivial continue indentation and unused argument warnings.

Yet there are still a large number of errors that seem like they need to be addressed:

ex.
filemanager/domain/index.py:35: [E1137(unsupported-assignment-operation), FileIndex.set] 'self.system' does not support item assignment
...
filemanager/domain/index.py:47: [E1135(unsupported-membership-test), FileIndex.contains] Value 'self.system' doesn't support membership test
...
filemanager/domain/index.py:60: [E1136(unsubscriptable-object), FileIndex.get] Value 'self.system' is unsubscriptable

Lots of these...

filemanager/domain/uploads/stored.py:101: [W0613(unused-argument), IStorageAdapter.get_full_path] Unused argument 'workspace'
filemanager/domain/uploads/stored.py:102: [W0613(unused-argument), IStorageAdapter.get_full_path] Unused argument 'u_file'
...

filemanager/domain/uploads/file_mutations.py:384: [W0201(attribute-defined-outside-init), SourceLog.__post_init__] Attribute '_logger' defined outside __init__
filemanager/domain/uploads/file_mutations.py:385: [W0201(attribute-defined-outside-init), SourceLog.__post_init__] Attribute '_f_path' defined outside __init__
filemanager/domain/uploads/file_mutations.py:386: [W0201(attribute-defined-outside-init), SourceLog.__post_init__] Attribute 'file_handler' defined outside __init__

Here at arXiv we aim for a Pylint target score of 9/10. :^) wink wink

I realize you are extremely busy. 

Would you like me to start eliminating the trivial sorts of Pylint errors? I'm not sure what's going on with so many unused argument errors and whether there is an easy global fix. Some of the other errors might be faster for you to resolve (if you think this is important).

DavidLFielding:
While I'm at it I ran '**pydocstyle** --convention=numpy --add-ignore=D401'

There are nearly 60 "Missing docstring" errors. Many of these are located in the newer file_mutations.py/storage.py code. Don't scream yet.

**mypy** generates a large number of errors. Most of these are trivial and easily correctable:

easy ones...lots of these...

filemanager/domain/tests/test_workspace.py:16: error: Function is missing a return type annotation
filemanager/domain/tests/test_workspace.py:16: note: Use "-> None" if function does not return a value

while some of these other errors might warrant a closer look:

filemanager/domain/uploads/stored.py:223: error: "IStorageAdapter" has no attribute "get_size_bytes"
filemanager/domain/uploads/stored.py:247: error: Returning Any from function declared to return "bytes"
filemanager/domain/uploads/translatable.py:77: error: "Type[datetime]" has no attribute "fromisoformat"
filemanager/domain/uploads/translatable.py:109: error: "TranslatableWorkspace" has no attribute "lastupload_readiness"
filemanager/domain/uploads/translatable.py:110: error: "TranslatableWorkspace" has no attribute "status"
...

along with several of the type mismatch flavor... 

filemanager/domain/uploads/checkpoint.py:307: error: Argument 1 to "_update_from_checkpoint" of "CheckpointWorkspace" has incompatible type "TranslatableWorkspace"; expected "CheckpointWorkspace"
...

You may want to take a quick look to see if any of these need to be addressed. 

Tomorrow I will fire up the submission UI suite and test the FM as I resume work on alpha tests.

filemanager/domain/index.py
erickpeirson: Removed files are handled separately (see line just above)

filemanager/domain/uploaded_file.py
erickpeirson: Slightly different from basename:

```python
>>> os.path.basename('foo/bar.txt')
'bar.txt'
>>> os.path.splitext('foo/bar.txt')[0]
'foo/bar'
```

> To be consistent should delete_file() become directory_sans_file()? 

Not sure what you mean -- this is just generating a string based on the path. delete_file() has side-effects (namely, deleting a file)

filemanager/domain/uploads/__init__.py
erickpeirson: > Does this implementation limit in any way the ability to provide a single-file submission with ancillary data files?

No, because ``.file_count`` excludes ancillary files, and ``.is_single_file_submission`` is just looking at ``.file_count``.

filemanager/domain/uploads/__init__.py
erickpeirson: Well, ``source_path`` is based on ``base_path``, which is where the system files live. So it should exist by extension of these being created. I guess if you wanted to store system files somewhere else; hadn't considered that scenario. Do you think we should support this?

filemanager/domain/uploads/file_mutations.py
erickpeirson: Agree it was misleading; fixed in 3d5ca4b

filemanager/process/check/tex_format.py
erickpeirson: These weren't implemented before the refactor, IIRC

filemanager/process/check/tex_generated.py
erickpeirson: fixed in 7545f3d

filemanager/process/check/unpack.py
erickpeirson: Dunno. This TODO was there before the refactor IIRC

filemanager/process/strategy.py
erickpeirson: cleaned up in 96c78e4

tests/test_api/_tests.py
erickpeirson: removed in f7d762c

tests/test_api/_tests.py
erickpeirson: Is this it? https://github.com/arXiv/arxiv-filemanager/blob/2f37ed6b42266a39d9aee99403c37f154d4bb824/tests/test_api/test_new_upload.py#L181

filemanager/domain/uploaded_file.py
DavidLFielding: I was really just poking fun at the method name. I apologize.

tests/test_api/test_ancillary_upload.py
erickpeirson: Created #47 since this seems like a significant issue

tests/test_api/test_checkpoints.py
erickpeirson: fixed in 734bf17

tests/test_api/test_eps_repair.py
erickpeirson: Yeah, looks like a lost track of this one. Per comment higher up, though, this was kind of a gnarly test. I'd like to see if we can set up a test that just exercise the EPS repair functionality itself -- there is a whole lot of unrelated API stuff here. Will create a ticket so this isn't lost.

erickpeirson: Created #48

filemanager/domain/uploads/__init__.py
DavidLFielding: Ahh. I saw new system directory and thought you were now sticking the system/checkpoint files there. My bad. I need to take another look at the code.

tests/test_api/test_existing_upload.py
erickpeirson: removed in 9c1da66

filemanager/domain/uploaded_file.py
erickpeirson: haha no worries :-)

filemanager/process/check/tex_format.py
DavidLFielding: Right. That was a little wishful thinking on my part.

tests/test_api/test_new_upload.py
erickpeirson: Makes sense. IIRC, I was seeing warnings disappear that seemed like they shouldn't disappear. Will create a ticket, as it would be good to take a broader look at errors/warnings (also per discussion in Slack)

erickpeirson: Created #49

tests/test_check_workspace.py
erickpeirson: That sounds like what was going on. I did a pretty deep dive on the contents of these files to verify that the expected behavior was occurring, so feeling pretty good about the size change based on what you've said

tests/test_api/_tests.py
DavidLFielding: I think these just migrated elsewhere. I seem to recall running into them elsewhere.

erickpeirson:
> Would you like me to start eliminating the trivial sorts of Pylint errors? I'm not sure what's going on with so many unused argument errors and whether there is an easy global fix. Some of the other errors might be faster for you to resolve (if you think this is important).

Good catches. Some of these are really things that need to be fixed, and some are not. 

For example, the ``unsupported-assignment-operation`` and other messages on those dicts are spurious, because pylint doesn't know that ``dataclasses`` is doing magic with ``field(...)``. So some of those can get ``# pylint: disable=``ed. I also added the following to the ``.pylintrc`` as blacklisted:
- ``unused-argument`` -- most of this comes from the use of protocols, and the standardized calling API of checker methods.
- ``duplicate-code`` -- this is largely coming from tests. May be a better way to do this, not an expert on pylint config.
- ``no-self-use`` -- this largely comes from the checkers; pylint doesn't know what patterns are being used here, so this is spurious.

erickpeirson:
> mypy generates a large number of errors. Most of these are trivial and easily correctable:

I'd recommend using ``./tests/type-check.sh filemanager`` rather than running mypy directly (maybe you are already doing this). I had a ``tee`` in the wrong spot in that script, so it was printing ignored errors (the ones in the tests) even though ultimately those aren't counted against the final result. That being said, I am still seeing errors (which is somewhat disheartening, because I really thought that I had fixed all of these):

```bash
$ ./tests/type-check.sh filemanager
filemanager/domain/uploads/stored.py:225: error: "IStorageAdapter" has no attribute "get_size_bytes"
filemanager/domain/uploads/stored.py:249: error: Returning Any from function declared to return "bytes"
filemanager/domain/uploads/translatable.py:77: error: "Type[datetime]" has no attribute "fromisoformat"
filemanager/domain/uploads/translatable.py:109: error: "TranslatableWorkspace" has no attribute "lastupload_readiness"
filemanager/domain/uploads/translatable.py:110: error: "TranslatableWorkspace" has no attribute "status"
filemanager/domain/uploads/translatable.py:111: error: "TranslatableWorkspace" has no attribute "lock_state"
filemanager/domain/uploads/translatable.py:112: error: "TranslatableWorkspace" has no attribute "source_type"
filemanager/domain/uploads/translatable.py:141: error: "Type[TranslatableWorkspace]" has no attribute "Status"
filemanager/domain/uploads/translatable.py:142: error: "Type[TranslatableWorkspace]" has no attribute "LockState"
filemanager/domain/uploads/translatable.py:143: error: "Type[TranslatableWorkspace]" has no attribute "SourceType"
filemanager/domain/uploads/translatable.py:152: error: "Type[TranslatableWorkspace]" has no attribute "Readiness"
filemanager/domain/uploads/checkpoint.py:307: error: Argument 1 to "_update_from_checkpoint" of "CheckpointWorkspace" has incompatible type "TranslatableWorkspace"; expected "CheckpointWorkspace"
filemanager/process/strategy.py:34: error: Function is missing a return type annotation
filemanager/process/strategy.py:34: note: Use "-> None" if function does not return a value
filemanager/process/strategy.py:52: error: Need type annotation for 'tasks'
filemanager/process/strategy.py:56: error: Function is missing a return type annotation
filemanager/process/strategy.py:61: error: Function is missing a return type annotation
filemanager/process/strategy.py:61: note: Use "-> None" if function does not return a value
filemanager/process/strategy.py:74: error: List comprehension has incompatible type List[UploadedFile]; expected List[Tuple[Any, ...]]
filemanager/process/strategy.py:75: error: Call to untyped function "await_completion" in typed context
filemanager/process/strategy.py:84: error: "IChecker" has no attribute "__iter__" (not iterable)
filemanager/process/strategy.py:95: error: No return value expected
filemanager/process/check/file_type.py:399: error: Argument 1 to "_type_of_latex2e" has incompatible type "UploadedFile"; expected "BytesIO"
filemanager/process/check/file_type.py:435: error: Incompatible types in assignment (expression has type "bytes", variable has type "str")
filemanager/process/check/file_type.py:464: error: Argument 1 to "_type_of_latex2e" has incompatible type "IO[Any]"; expected "BytesIO"
filemanager/process/check/cleanup.py:171: error: Argument 1 to "findall" of "Pattern" has incompatible type "mmap"; expected "bytes"
filemanager/process/check/cleanup.py:226: error: Argument 1 to "search" of "Pattern" has incompatible type "str"; expected "bytes"
filemanager/process/check/cleanup.py:233: error: Argument 1 to "search" of "Pattern" has incompatible type "str"; expected "bytes"
filemanager/process/check/cleanup.py:239: error: Argument 1 to "search" of "Pattern" has incompatible type "str"; expected "bytes"
filemanager/process/check/cleanup.py:247: error: Argument 1 to "search" of "Pattern" has incompatible type "str"; expected "bytes"
filemanager/process/check/cleanup.py:253: error: Unsupported operand types for + ("bytes" and "str")
filemanager/process/check/cleanup.py:256: error: Argument 1 to "search" of "Pattern" has incompatible type "str"; expected "bytes"
filemanager/controllers/checkpoint.py:122: error: "print_exc" does not return a value
filemanager/controllers/checkpoint.py:233: error: "print_exc" does not return a value
filemanager/controllers/checkpoint.py:282: error: "print_exc" does not return a value
filemanager/controllers/upload.py:101: error: Incompatible types in assignment (expression has type "None", variable has type "UploadWorkspace")
filemanager/controllers/upload.py:177: error: Argument 1 to "create" of "FileMutationsWorkspace" has incompatible type "Union[str, None, Any]"; expected "str"
filemanager/controllers/files.py:221: error: "UploadFileSecurityError" has no attribute "description"
```

PR #45, opened 2019-07-22
ARXIVNG-2053: TeX-Produced check

DavidLFielding:
This PR contains code routines and tests for detecting TeX Produced Postscript and PDF files. There is one routine to handle Postscript and another for PDF.

Due to the refactor and evolving code organization I only wrote the core routines with tests and did not wire this up in the old file manager. These routines/tests need to be added to refactored File Manager and called from the newly organized test suite.

pdfinfo/pdffonts:

TeX Produced check uses pdfinfo and pdffonts commands for extracting metadata from files. This will need to be installed in File Manager Docker image.

    yum install poppler-utils

Integration hints:

The TeX Produced check is called ln the Legacy system's Source.pm process routine just before we perform type-specific processing (compile TeX source submission).

TeX Produced is only called on submissions of type PS or PDF. Since some of these submission types allow multiple files the tex produced routine is called for all files contained in the submission (which will normally be a single file PDF/Postsc).

process:

$self->check_tex_produced if ( $self->sub_type eq 'ps' || $self->sub_type eq 'pdf' );

check: This routine is not completely implemented in the Python TeXProduced package since this 
            might make more sense integrating into the upload checks logic.

   foreach my $file ( @{$self->files} ) {
	  if (($file->type eq 'postscript' && tex_produced_ps($file->filepath))
	      or ($file->type eq 'pdf' && tex_produced_pdf($file->filepath))) {
            $self->tex_produced($file); # Generates/registers errors
	  }
  }


DavidLFielding:
Ready for integration. I'm not clear on travis failures but I suspect this was set up to check refactored file manager.

DavidLFielding:
This functionality has been moved to the Submission UI and Core repositories.

PR #46, opened 2019-08-13
ARXIVNG-2032: Cleaned up some pydocstyle errors from checkpoint implementation.

DavidLFielding:
I noticed a few minor pydoc issues in checkpoint implementation and fixed them. Minor.


erickpeirson (approved):
Looks good; thanks for catching this

PR #50, opened 2019-08-17
Errors and warnings

erickpeirson:
This makes some fixes to handling of errors (and warnings):

- Errors are identified with a ``code`` -- a file may only have one error for a particular code, and a workspace may only have one non-file error for a particular code.
- Removing a file no longer automatically generates an error/warning. The caller must separately generate an error/warning. This gives a bit more flexibility, and better separation of concerns.
- Added support for removing errors.
- General cleanup/tidying of error-related methods and properties.

### DIFF
--- a/filemanager/domain/uploads/errors_and_warnings.py
+++ b/filemanager/domain/uploads/errors_and_warnings.py
@@ -1,12 +1,12 @@
 """Provides :class:`.ErrorsAndWarnings`."""
 
-from typing import List, Optional, Any, Iterator, Dict, Callable, cast
+from typing import List, Optional, Any, Iterator, Dict, Callable, Tuple, cast
 
 from dataclasses import dataclass, field
 from typing_extensions import Protocol
 
 from ..uploaded_file import UserFile
-from ..error import Error
+from ..error import Error, Code, Severity
 from ..index import FileIndex
 from .base import IBaseWorkspace
 
@@ -55,21 +55,21 @@ class IErrorsAndWarnings(Protocol):
     def warnings_active(self) -> List[Error]:
         """Warnings for active files only."""
 
-    def add_error(self, u_file: UserFile, msg: str,
-                  severity: Error.Severity = Error.Severity.FATAL,
+    def add_error(self, u_file: UserFile, code: Code, msg: str,
+                  severity: Severity = Severity.FATAL,
                   is_persistant: bool = True) -> None:
         """Add an error for a specific file."""
 
-    def add_error_non_file(self, msg: str,
-                           severity: Error.Severity = Error.Severity.FATAL,
+    def add_error_non_file(self, code: Code, msg: str,
+                           severity: Severity = Severity.FATAL,
                            is_persistant: bool = True) -> None:
         """Add an error for the workspace that is not specific to a file."""
 
-    def add_warning(self, u_file: UserFile, msg: str,
+    def add_warning(self, u_file: UserFile, code: Code, msg: str,
                     is_persistant: bool = True) -> None:
         """Add a warning for a specific file."""
 
-    def add_warning_non_file(self, msg: str,
+    def add_warning_non_file(self, code: Code, msg: str,
                              is_persistant: bool = False) -> None:
         """Add a warning for the workspace that is not specific to a file."""
 
@@ -79,16 +79,21 @@ class IErrorsAndWarnings(Protocol):
                      is_removed: bool = False) -> List[str]:
         """Get all warnings for the file at ``path``."""
 
+    def remove_error(self, code: Code, path: Optional[str] = None) -> None:
+        """Remove an error."""
+
 
 class IErrorsAndWarningsWorkspace(IWorkspace, IErrorsAndWarnings, Protocol):
     """Interface for workspace with errors and warnings behavior."""
 
+Path = Optional[str]
 
 @dataclass
 class ErrorsAndWarnings(IErrorsAndWarnings):
     """Adds methods for handling errors and warnings."""
 
-    _errors: List[Error] = field(default_factory=list)
+    # _errors: List[Error] = field(default_factory=list)
+    _errors: Dict[Tuple[Path, Code], Error] = field(default_factory=dict)
 
     __internal_api = None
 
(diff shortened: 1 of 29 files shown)


DavidLFielding:
Could you expand more on the necessity of codes and their use in the submission UI and file manager. It sounds like removing a file is a special case that doesn't automatically generate an error. If there is no error generated how does the client (submission UI) decide what to do? 

I'm just trying to better understand the justification for these changes. From the UI's perspective all it really cares about are two things: 1) a warning is displayed and ignorable 2) an error is displayed and stops progress in the workflow. Warnings and errors are displayed with active files or at the top of the page. This seemed perfectly adequate and a very simple solution. When I saw the extra messages in the UI during testing I thought some of them needed to be directed to the log instead of to the client. I believe this would have eliminated the flood of too many messages.

Does the ability to remove an error basically silence an error even though the condition causing the error still persists?

I still need to look over the code more closely. I was just surprised by the size of the PR given the problem it addresses.

erickpeirson:
> Could you expand more on the necessity of codes and their use in the submission UI and file manager.

It's just a name for an error. Reduces ambiguity, and is more useful for consumers of the file manager API in general. Most importantly, it allows us to remove errors programmatically. Since we're not running all of the checks on all of the files on each operation, it's important to be able to perform operations on errors and warnings.

> It sounds like removing a file is a special case that doesn't automatically generate an error. If there is no error generated how does the client (submission UI) decide what to do?

If removing a file should be brought to the user or client's attention, then a warning or error should also be generated. Part of the diff is making sure that this happens. The change here is that they aren't intrinsically linked.

> When I saw the extra messages in the UI during testing I thought some of them needed to be directed to the log instead of to the client. I believe this would have eliminated the flood of too many messages.

There were several things going on. 

One was that stale errors were not getting removed (e.g. invalid source type). One of the changes in the refactor was that checks are applied iteratively until there are no more files to check. Unpacking tar or zip files is a check that generates more files, so this results in multiple passes of the source type check. Even though the source was (say TeX) ultimately valid, the initial pass generated an invalid source error. So we needed a way to clear the prior error when it was clear that the source was valid 

Another thing that was happening was that error generation and logging got a little goofy during the refactor. Just cleaning things up so that we have a clear API, and it's clear where errors/warnings are generated, helped ungoof things.

> Does the ability to remove an error basically silence an error even though the condition causing the error still persists?

The pattern currently is that a checker adds and removes errors. The checker is responsible for identifying adverse conditions, so this seems appropriate. Sure, a checker could lie to the client/user, but that wouldn't be very nice.

> I was just surprised by the size of the PR given the problem it addresses.

The size of the PR stems mainly from the fact that errors and warnings are generated in quite a few places, and i changed the signature for adding errors to include a code. Plus some convenience properties.

DavidLFielding:
Your detailed clarification above helps a lot. Thanks.

filemanager/domain/uploads/errors_and_warnings.py
```diff
@@ -55,40 +55,51 @@ def warnings(self) -> List[Error]:
...
         """Get all warnings for the file at ``path``."""
 
+    def remove_error(self, code: Code, path: Optional[str] = None) -> None:
+        """Remove an error."""
+
 
 class IErrorsAndWarningsWorkspace(IWorkspace, IErrorsAndWarnings, Protocol):
     """Interface for workspace with errors and warnings behavior."""
 
+Path = Optional[str]
```
DavidLFielding: This line seems a little odd to me. Doesn't appear to be used. Maybe this is related to the use of protocols.

filemanager/domain/uploads/tests/test_errors_and_warnings.py
```diff
@@ -0,0 +1,100 @@
...
+        eaw.add_error(u_file, 'foo_error', 'this is also a foo error')
+
+        self.assertTrue(eaw.has_errors)
+        self.assertTrue(eaw.has_errors_fatal)
+        self.assertEqual(len(eaw.errors), 1, 'Only one error per code')
+
+        eaw.add_error(u_file, 'baz_error', 'this is not a foo error')
+        self.assertEqual(len(eaw.errors), 2)
+
+    def test_remove_error_file(self):
```
DavidLFielding, line 44: So if I do add_error 'foo_error', add_error 'foo_error', add_error 'foo_error', and remove_error 'foo_error' I have 0 errors. Interesting. I'd need to revisit the specific low-level errors before I could determine there is a problem. Likely fine.

filemanager/domain/uploads/tests/test_errors_and_warnings.py
```diff
@@ -0,0 +1,100 @@
...
+
+        self.assertTrue(eaw.has_errors)
+        self.assertTrue(eaw.has_errors_fatal)
+        self.assertEqual(len(eaw.errors), 1)
+
+    def test_remove_error_file_from_file(self):
+        """Remove an error for a file from the file itself."""
+        api = mock.MagicMock(spec=IWorkspace, files=FileIndex())
+        u_file = UserFile(api, 'foo/path', 42)
+        api.files.set(u_file.path, u_file)
```
DavidLFielding, line 82: I've seen quite a few api.files.set(u_file.path, u_file). Isn't u_file alone sufficient for the set method to determine path. I haven't looked at other usages so this is likely fine.

filemanager/process/check/cleanup.py
```diff
@@ -72,18 +91,28 @@ def check(self, workspace: Workspace, u_file: UserFile) \
 class RepairDOSEPSFiles(BaseChecker):
     """Repair DOS EPS files."""
 
+    FAILED_STRIP_PREVIEW: Code = 'strip_preview_failed'
```
DavidLFielding, line 94: Such similar names yet different types. Not confusing as long as you know MESSAGE indicates str type, otherwise you have a Code. Not sure I'd go as far as LEADING_PREVIEW_STRIPPED_CODE. Just noting but not sure it really matters much.

DavidLFielding (approved):
I scanned through the warning/errors changes and didn't see anything terrible. Looks fine. I'll do some live testing as soon as I get a chance.

filemanager/domain/uploads/tests/test_errors_and_warnings.py
erickpeirson: That is correct; adding multiple instances of the same error is redundant. If two instances of an error need to be added and removed independently, then they should simply be different errors (with different codes).

filemanager/domain/uploads/tests/test_errors_and_warnings.py
erickpeirson: Yes; I added the method ``add(u_file: UserFile)`` to ``FileIndex`` in this PR, and we can start using that instead.

erickpeirson:
> didn't see anything terrible

Good to hear ;-)

filemanager/domain/uploads/tests/test_errors_and_warnings.py
DavidLFielding: I thought different errors were being mapped to the same code. My bad. All is good.

PR #51, opened 2020-02-21
Workspace source format not cleared on "delete_all"

bdc34:
This bug appears to be a simple oversight to clear the source_type.


DavidLFielding (approved):
This change looks good and makes sense to me. I did not actually try it at this time. Seems straightforward and added test makes sure format is cleared.

mhl10 (approved):
looks good

